PR #4, opened 2024-04-24
Add runtime framework for managing subscriptions and update parsing.

ray-kast:
Currently this branch depends on a local fork of the `topograph` crate, since I am [updating it](https://github.com/ray-kast/topograph/compare/master...v0.4-wip) in parallel to act as an async data-parellel concurrency limiter, flow controller, and push-pull adapter.

### DIFF
--- a/crates/runtime/src/buffer.rs
+++ b/crates/runtime/src/buffer.rs
@@ -5,10 +5,14 @@ use topograph::{
     executor::{Executor, Nonblock, Tokio},
     prelude::*,
 };
-use tracing::error;
-use yellowstone_grpc_proto::{geyser::SubscribeUpdate, tonic::Status};
+use tracing::{error, warn};
+use yellowstone_grpc_proto::{
+    geyser::{subscribe_update::UpdateOneof, SubscribeUpdate},
+    tonic::Status,
+};
+use yellowstone_vixen_core::{AccountUpdate, TransactionUpdate};
 
-use crate::{parser::Message, parser_manager::ParserManager, yellowstone, Parser};
+use crate::{handler::DynHandlerPack, yellowstone, HandlerManagers};
 
 #[derive(Default, Debug, clap::Args, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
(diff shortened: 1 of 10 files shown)


crates/runtime/src/buffer.rs
```diff
@@ -0,0 +1,46 @@
+use std::pin::pin;
+
+use futures_util::{Stream, StreamExt};
+use topograph::{
+    executor::{Executor, Nonblock, Tokio},
+    prelude::*,
+};
```
kespinola, line 7: Really like that you are building the thread manager into a crate. So many use cases for managing pool of worker threads.

crates/runtime/src/buffer.rs
```diff
@@ -0,0 +1,46 @@
...
+    client: yellowstone::YellowstoneStream<I, T, S>,
+) -> Buffer {
+    let BufferOpts { jobs } = opts;
+
+    let exec = Executor::builder(Nonblock(Tokio))
+        .num_threads(jobs)
+        .build(|j, h| async move {
+            tracing::info!(?j);
+        })
+        .unwrap_or_else(|i| match i {});
```
kespinola, line 59: is this the callback for the worker? as in the input of `u` on LN37.

crates/test/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    let Opts { config } = Opts::parse();
+    let config = std::fs::read_to_string(config).expect("Error reading config file");
+    let config = toml::from_str(&config).expect("Error parsing config");
+    vixen::run(
+        config,
+        vixen::ParserManager::<vixen::BoxedParser>::new(
+            [("my-parser".into(), Box::new(Parser) as vixen::BoxedParser)]
+                .into_iter()
+                .collect(),
+        ),
```
kespinola: If parsers set their program address then that could be used to decide which parser to use. 

For accounts the owner field is the program.

On transactions each instruction is aware of the program. We haven't looked at transactions on this go but the transaction flattens out all instructions but tracks its stack height so we can group by stack height and process up the stack height referencing the owner program to pick which parser to apply.

crates/test/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    async fn process_transaction<'a>(
+        &'a self,
+        txn: &'a TransactionUpdate,
+    ) -> Result<(), Self::Error> {
+        info!(?txn);
+        Ok(())
+    }
+}
```
kespinola: The parser should operate on a per instruction level. What the runtime should do is parse the transaction and run through the correct instruction. The runtime should ensure instructions for a transaction are executed in correct order if required.

PR #13, opened 2024-07-24
Token Parsers + Metrics support 

kespinola:
### Changes
- Add crate which uses feature flags to include certain program parsers.
- Parser crate has optional `mock` feature for including spec framework for testing parsers. It generates fixtures based on configured network which requests the asset, saves it to the project, loads fixture into spec for testing the parser. 
- Add metrics support (opentelemetry +Prometheus) to get metrics on accounts and transactions parsing 

```rust
#[test]
fn mint_parse_success() {
  let account = account_mock!("6dhTynDkYsVM7cbF7TKfC9DWB636TcEM935fq7JzL2ES");
  
  let mint = parse_update!(parser, account);

  assert_eq!(mint.decimals, 9);
}
```

### Parsers
- [x] Token Program
- [x] Associated Token Program
- [x] Token Extensions


crates/parser/src/token_extensions.rs
```diff
@@ -0,0 +1,182 @@
...
+    TokenAccount,
+    Mint,
+    Multisig,
+}
+
+#[derive(Debug)]
+pub struct ExtensionData {
+    pub extension: ExtensionType,
+    pub extension_data: Vec<u8>,
+}
```
kespinola: Lets represent extension data as an enum
```suggestion
pub enum ExtensionData {
   ImmutableOwner(ImmutableOwnerData),
}
```

crates/parser/src/token_extensions.rs
```diff
@@ -0,0 +1,182 @@
...
+pub struct ExtendedMint {
+    pub base_account: Mint,
+    pub extension_data_vec: Vec<ExtensionData>,
+}
+
+impl ExtendedMint {
+    fn from_account_data(data_bytes: &[u8]) -> Result<ExtendedMint, ProgramError> {
+        let unpacked = StateWithExtensions::<Mint>::unpack(data_bytes)?;
+        let extension_types = unpacked.get_extension_types()?;
+        let mut extension_data_vec: Vec<ExtensionData> = vec![];
```
kespinola: Can set with_capacity since you know the extension_types

crates/parser/src/token_extensions.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+#[derive(Debug)]
+pub struct ExtendedTokenAccount {
+    pub base_account: Account,
+    pub extension_data_vec: Vec<ExtensionData>,
+}
+
+impl ExtendedTokenAccount {
+    fn from_account_data(data_bytes: &[u8]) -> Result<ExtendedTokenAccount, ProgramError> {
```
kespinola:
```suggestion
    fn try_from_account_data(data_bytes: &[u8]) -> Result<ExtendedTokenAccount, ProgramError> {
```
When return a result its a `try_from`

crates/parser/src/token_extensions.rs
```diff
@@ -0,0 +1,182 @@
...
+
+                assert_eq!(parsed_data.is_ok(), true);
+
+                let parsed_data = parsed_data.unwrap();
+
+                println!("Parsed Transfer Hook Account Extension: {:?}", parsed_data);
+            }
+            _ => panic!("Invalid account type"),
+        }
+    }
```
kespinola: Why are these 2 separate unpack operations? Can everything be embedded into the top level `TokenExtensionState`?

crates/parser/src/token_extensions.rs
```diff
@@ -0,0 +1,182 @@
...
+        constants::token_program_constants::TOKEN_ACCOUNT_WITH_EXTENSION,
+        token_extension_data_parsers::parse_transfer_hook_account_extension,
+    };
+
+    use super::*;
+
+    #[test]
+    fn test_token_account_parsing() {
+        let token_account = TokenExtensionState::try_unpack(TOKEN_ACCOUNT_WITH_EXTENSION);
+        assert_eq!(token_account.is_ok(), true);
```
kespinola:
```suggestion
```

The unwrap covers the same case.

crates/parser/src/token_extensions.rs
```diff
@@ -0,0 +1,182 @@
...
+        let token_account = token_account.unwrap();
+        match token_account {
+            TokenExtensionState::ExtendedTokenAccount(ext_token_account) => {
+                println!("Token Account with Extensions: {:?}", ext_token_account);
+                let ext_data = ext_token_account.extension_data_vec;
+                assert_eq!(ext_data.len(), 1);
+                assert_eq!(ext_data[0].extension, ExtensionType::TransferHookAccount);
```
kespinola:
```suggestion
                assert_eq!(ext_data[0], ExtensionData::TransferHookAccount(_)); 
```

If enum is used can drop the extension field

crates/parser/src/token_extension_helpers.rs
```diff
@@ -0,0 +1,191 @@
+use spl_token_2022::{
+    extension::{self, BaseStateWithExtensions, ExtensionType, StateWithExtensions},
+    solana_program::program_error::ProgramError,
+    state::{Account, Mint},
+};
+// #[cfg(feature="extension-data-parsers")]
+use spl_type_length_value::variable_len_pack::VariableLenPack;
+// #[cfg(feature="extension-data-parsers")]
```
kespinola: Include if `token-extensions` is included no need for other feature flag

crates/parser/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "yellowstone-vixen-parser"
+description = "Vixen program parsers for the solana program library."
+version = "0.0.0"
+edition = "2021"
+license = "AGPL-3.0-or-later"
```
kespinola: should be MIT license

crates/mock/src/async_client.rs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        ]
+    });
+
+    let mut response = client
+        .post(rpc_url)
+        .body(request_body.to_string())
+        .header("Content-Type", "application/json")
+        .send()
+        .await?;
```
kespinola: Was there an issue with using solana_sdk?

crates/mock/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+            Some(account) => Some(get_subscribe_update_account(pubkey, account)),
+            None => None,
+        }
+    } else {
+        let account = fetch_account_data_from_file(pubkey, cluster);
+        match account {
+            Some(account) => Some(get_subscribe_update_account(pubkey, account)),
+            None => None,
+        }
+    }
```
kespinola: swap order so the truthy case is first. easier for brain to process.

crates/mock/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+    use solana_sdk::genesis_config::ClusterType;
+    use utils::test_parsing_token_extension_program;
+
+    use super::*;
+    #[tokio::test]
+    async fn mock() {
+        let account = load_account_fixtures(VIXEN_TEST_MINT_DEVNET, ClusterType::Devnet).await;
+        assert!(account.is_some(), "Account not found");
+        let account = account.unwrap();
+        test_parsing_token_extension_program(account).await;
```
kespinola: Mock library doesn't need to test token extension parsing just the account.

crates/mock/src/utils.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub data: Vec<String>,
+    pub executable: bool,
+    pub lamports: u64,
+    pub owner: String,
+    #[serde(rename = "rentEpoch")]
+    pub rent_epoch: u64,
+    pub space: u64,
+}
+
+pub async fn fetch_account_sync(pubkey: &Pubkey, cluster: ClusterType) -> Option<Account> {
```
kespinola: doesn't need to be async 
```suggestion
pub fn fetch_account_sync(pubkey: &Pubkey, cluster: ClusterType) -> Option<Account> {
```

crates/mock/src/utils.rs
```diff
@@ -0,0 +1,238 @@
...
+pub async fn test_custom_parser(
+    sub_account_update: SubscribeUpdateAccount,
+    parser: impl Parser<Input = AccountUpdate>,
+) {
+    let data = parser.parse(&sub_account_update).await;
+
+    assert!(data.is_ok(), "Error parsing account");
+
+    let data = data.unwrap();
+}
```
kespinola: drop these.

crates/mock/src/utils.rs
```diff
@@ -0,0 +1,238 @@
...
+    let account_data = account_info.data.first().unwrap();
+    let account_data = general_purpose::STANDARD.decode(account_data).unwrap();
+    Account {
+        lamports: account_info.lamports,
+        data: account_data,
+        owner: Pubkey::from_str(&account_info.owner).unwrap(),
+        executable: account_info.executable,
+        rent_epoch: account_info.rent_epoch,
+    }
+}
```
kespinola: This should be a TryFrom
```
impl TryFrom<AccountInfo> -> Account {}
```

crates/mock/src/utils.rs
```diff
@@ -0,0 +1,238 @@
...
+pub fn read_from_file(file_path: &str) -> Option<AccountInfo> {
+    let file = std::fs::read(file_path);
+    match file {
+        Ok(data) => {
+            let account: AccountInfo = serde_json::from_slice(&data).unwrap();
+            Some(account)
+        }
+        Err(_) => None,
+    }
+}
```
kespinola: Look at using map and ok
```suggestion
pub fn read_from_file(file_path: &str) -> Option<AccountInfo> {
    std::fs::read(file_path)
      .map(|data| {
        serde_json::from_slice::<AccountInfo>(&data)
      }).ok()
}
```

crates/mock/src/utils.rs
```diff
@@ -0,0 +1,238 @@
...
+
+pub fn fetch_account_data_from_file(pubkey: &str, cluster: ClusterType) -> Option<AccountInfo> {
+    let account_file_path = get_account_data_file_path(pubkey, cluster);
+    match account_file_path {
+        Some(path) => read_from_file(&path),
+        None => None,
+    }
+}
```
kespinola: same but just map here

most match statement can be represented as using method chaining.

crates/mock/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+#[cfg(test)]
+mod tests {
+    const VIXEN_TEST_MINT_DEVNET: &str = "AZLFB7QYN8oZK8wfUt65feHFjDonhiPHQGDcoWfZDPFf";
+    use solana_sdk::genesis_config::ClusterType;
+    use utils::test_parsing_token_extension_program;
+
+    use super::*;
+    #[tokio::test]
+    async fn mock() {
+        let account = load_account_fixtures(VIXEN_TEST_MINT_DEVNET, ClusterType::Devnet).await;
```
kespinola: @ray-kast do you have recommendations on loading up the clustertype from config so it doesn't need to be passed in each call and should we make this blocking?

```
let account = account_fixture!("1234...xyz")
```

This is my ideal API.

crates/parser/src/token_extension_helpers.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub fn parse_immutable_owner_extension(data_bytes: &[u8]) -> Result<ImmutableOwner,ProgramError> {
+        let immutable_owner = pod_from_bytes::<extension::immutable_owner::ImmutableOwner>(data_bytes)?;
+        Ok(immutable_owner.to_owned())
+    }
+
+
+    pub fn parse_transfer_fee_amount_extension(data_bytes: &[u8]) -> Result<extension::transfer_fee::TransferFeeAmount,ProgramError> {
+        let transfer_fee_amount = pod_from_bytes::<extension::transfer_fee::TransferFeeAmount>(data_bytes)?;
+        Ok(transfer_fee_amount.to_owned())
+    }
```
kespinola: You may be able to use generics to consildated this

```rust
pub fn parse_extension_data::<E: >(data_bytes: &[u8]) -> Result<E, ProgramError> {
  pod_from_byts::<E>(data_bytes).map_err(Into::into)
}
```

crates/mock/src/async_client.rs
Nagaprasadvr: yeah there is no async getAccountInfo call

crates/parser/src/token_extensions.rs
Nagaprasadvr: this errors so using match for this

crates/mock/src/async_client.rs
```diff
@@ -0,0 +1,60 @@
...
+pub async fn fetch_account_async(
+    cluster: ClusterType,
+    account_pubkey: &str,
+) -> Result<AccountInfo, Box<dyn std::error::Error>> {
+    let rpc_url = match cluster {
+        ClusterType::MainnetBeta => "https://api.mainnet-beta.solana.com".to_string(),
+        ClusterType::Devnet => "https://api.devnet.solana.com".to_string().to_string(),
+        ClusterType::Testnet => "https://api.testnet.solana.com".to_string(),
+        _ => "https://api.mainnet-beta.solana.com".to_string(),
+    };
```
kespinola: Thinking on this I'd like the this to be read off config since larger teams will have private RPC. It could be a follow up and something for you to think about @ray-kast related to Vixen config.

crates/mock/src/utils.rs
```diff
@@ -0,0 +1,182 @@
...
+use solana_client::rpc_client::RpcClient;
+use solana_sdk::pubkey::Pubkey;
+use solana_sdk::{account::Account, genesis_config::ClusterType};
+use yellowstone_grpc_proto::geyser::{SubscribeUpdateAccount, SubscribeUpdateAccountInfo};
+use yellowstone_vixen_core::Parser;
+use yellowstone_vixen_parser::{TokenProgramParser, TokenProgramState};
+
+use crate::async_client::fetch_account_async;
+
+const FIXTURES_PATH: &str = "./fixtures";
```
kespinola: Another spot to be configurable @ray-kast

crates/parser/README.md
```diff
@@ -0,0 +1,34 @@
+# Yellowstone Vixen Parser
+
+This crate provides several account parsers, such as Token and TokenExtension. These parsers can be imported from this crate and used within yellowstone-vixen.
+
+## Example
+
```
kespinola, line 14: should show the imports as well.

And include how to install.

crates/mock/src/async_client.rs
Nagaprasadvr: Make sense we can ask users to create a .env or something or else toml also works

crates/mock/src/utils.rs
Nagaprasadvr: i feel we can add it to .env file itself wyt?

ray-kast (changes requested):
I've rebased the branch onto some work that I've done that can also be merged along with this.  The API for the runtime crate has changed, it looks like this breaks the README and possibly some other files.

We probably want to avoid dotenv for config, it would be preferable to offer a configuration object that can be deserialized from something like TOML.

Nagaprasadvr:
> has changed, it looks like this breaks the README and possibly some oth

have fixed readme

PR #15, opened 2024-08-01
feat : export metrics data ( opentelemetry  + Prometheus)

Nagaprasadvr:
1. Export opentelemetry and Prometheus metrics data 
2. Setup docker config for running opentelemetry and Prometheus services which will pull this metrics data and process it


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 members = ["crates/*"]
-exclude = ["crates/test", "crates/test-streams"]
+exclude = ["crates/demo", "crates/test-streams"]
```
kespinola: @ray-kast what do you think about have our examples in `examples`?

```
examples/
  simple
  streams
crates/
  core
  runtime
```

```suggestion
exclude = ["examples/*"]
```

Nagaprasadvr: yeah this is ideal makes sense

kespinola: @Nagaprasadvr make this change please

Nagaprasadvr: Will do

Nagaprasadvr:
#15 check this

Issue #3, opened 2024-04-22
Transaction Transforming V0


kespinola:
### Issue
Solana transaction objects are optimized for size but when it time to index them its difficult to assemble the instructions into their logic structure.

### References
https://docs.rs/solana-transaction-status/1.18.11/solana_transaction_status/struct.TransactionStatusMeta.html

PR #16, opened 2024-08-07
Setup Ixs Parsers (Token program and Token extensions program)

Nagaprasadvr:
WIP
Issue #3 

- [ ] Setup Transaction Parsers to Parse tx data 
- [ ] make Instruction data (including cpi calls , ix name , params ), account pubkeys , human readable
- [ ]  Start with Token Program parsers and eventually parse Other Program Ixs
- [ ]  Add Tests for Token Extensions account parsing using mock


crates/parser/src/ix_parser/token_program/token_ix_context.rs
```diff
@@ -0,0 +1,148 @@
...
+    ThawAccount(ThawAccount),
+    TransferChecked(TransferChecked),
+    ApproveChecked(ApproveChecked),
+    MintToChecked(MintToChecked),
+    BurnChecked(BurnChecked),
+    SyncNative(SyncNative),
+    GetAccountDataSize(GetAccountDataSize),
+    InitializeImmutableOwner(InitializeImmutableOwner),
+    AmountToUiAmount(AmountToUiAmount),
+    UiAmountToAmount(UiAmountToAmount),
```
kespinola: What are these instructions and why not use the structs provided by the program?

crates/parser/src/ix_parser/token_program/token_program_parser.rs
```diff
@@ -0,0 +1,311 @@
...
+        match ix_type {
+            TokenInstruction::Transfer { amount } => {
+                let source = ix_data.accounts[0];
+                let destination = ix_data.accounts[1];
+                Ok(TokenIxContext::Transfer(Transfer {
+                    source,
+                    destination,
+                    amount,
+                }))
+            },
```
kespinola: Accounts should be in another struct. We have TokenInstruction as the data and then accounts being a context struct with the accounts.

examples/simple/src/main.rs
```diff
@@ -46,13 +44,18 @@ fn main() {
...
+            //     handler::boxed(vixen::HandlerPack::new(TokenExtensionProgramParser, [
+            //         Handler,
+            //     ])),
+            //     handler::boxed(vixen::HandlerPack::new(TokenProgramParser, [Handler])),
+            // ]),
+            HandlerManager::empty(),
+            transaction: HandlerManager::new([handler::boxed(vixen::HandlerPack::new(
+                TokenProgramIxsParser,
+                [Handler],
+            ))]),
```
kespinola: @ray-kast program's are only aware of instructions so we either need to make this instructions and then vixen needs to decompose the transaction and invoke instruction parsers and handlers. 

Other option is register all instruction parsers and then a single handler for the transaction. 

I think 1 is easiest for the integrator but using full transaction may support some use case by giving the integrator the whole transaction vs just pieces of it.

kespinola (changes requested):

crates/parser/src/ix_parser/token_program/token_ix_context.rs
Nagaprasadvr: these ixs are from spl token program itself

crates/parser/src/ix_parser/token_program/token_program_parser.rs
Nagaprasadvr: image
something like this?

crates/parser/fixtures/BtS..uyu.json
```diff
@@ -0,0 +1,38 @@
+{
```
kespinola: Oh the filename should be the whole pubkey.

crates/parser/src/ix_parser/token_extensions/token_extension_parser.rs
```diff
@@ -0,0 +1,117 @@
...
+    vixen_ix::structure::{InstructionParser, InstructionUpdate},
+};
+
+#[derive(Debug)]
+pub struct TokenExtensionProgramIxParser;
+
+impl InstructionParser<TokenExtensionProgramIx> for TokenExtensionProgramIxParser {
+    fn parse_readable_ix(ix_update: &InstructionUpdate) -> Result<TokenExtensionProgramIx, String> {
+        match TokenInstruction::unpack(&ix_update.data) {
+            Ok(ix) => match ix {
```
kespinola: This is looking good. Nj. 
```suggestion
impl InstructionParser<TokenExtensionProgramIx> for TokenExtensionProgramIxParser {
    fn parse(ix_update: &InstructionUpdate) -> Result<TokenExtensionProgramIx, String> {
        match TokenInstruction::unpack(&ix_update.data) {
            Ok(ix) => match ix {
```

crates/parser/src/ix_parser/vixen_ix/structure.rs
```diff
@@ -0,0 +1,323 @@
+use serde_json::Value;
```
kespinola: The transaction decompose mat be better spot for core or really where the similar logic for accounts resides.

crates/parser/src/ix_parser/token_extensions/token_extension_parser.rs
Nagaprasadvr: renaming to parse_ix bcos we already have parse fn inside Parser trait it will cause fn name conflict

crates/parser/src/ix_parser/vixen_ix/structure.rs
Nagaprasadvr: moved to core

ray-kast (changes requested):
I have some architecture decisions I need to make about the runtime and core crates regarding how this should be integrated.  I can handle that once I'm done working on streams.

Nagaprasadvr:
> I have some architecture decisions I need to make about the runtime and core crates regarding how this should be integrated. I can handle that once I'm done working on streams.

Cool , I will move on to mock setup and leave Ix parser logic as it is ,

ray-kast:
Superseded by #19.

PR #18, opened 2024-08-24
Implement gRPC program streaming service, plus ancillary refactors

ray-kast:
Commits:
- **Retool metrics, reinstate OTel, bump dependencies.**
- **Add example crate for basic OpenTelemetry metrics.**
- **WIP: Implement gRPC streams + refactor everything.**

Remaining tasks:
- [ ] Test new push-based metrics with Prometheus gateway
- [ ] Test updated runtime code with multiple parsers
- [ ] Update all documentation to reflect API changes
- [ ] Solicit feedback on gRPC implementation
- [ ] Resolve remaining TODOs
- [ ] Consolidate parts of the public-facing API that are no longer used
- [ ] Improve performance and precision of `InstructionPipeline` dispatch
- [ ] Implement stubbed-out additions to `yellowstone-vixen-parser`
- [ ] Perform integration testing on streams and instruction parsers


crates/core/src/lib.rs
```diff
@@ -104,6 +140,26 @@ impl TryFrom<&[u8]> for Pubkey {
...
 
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+pub enum PubkeyFromStrError {
+    #[error("Invalid base58 string")]
+    Bs58(#[from] bs58::decode::Error),
+    #[error("Invalid key length, must be 32 bytes")]
+    Len(#[from] std::array::TryFromSliceError),
+}
+
+impl FromStr for Pubkey {
```
Nagaprasadvr, line 190: helpful

crates/runtime/src/buffer.rs
```diff
@@ -1,140 +1,162 @@
...
+    }
+}
+
+struct Job(tracing::Span, SubscribeUpdate);
+
+struct Handler<M: Instrumenter> {
+    pipelines: Arc<PipelineSets>,
+    counters: Arc<Counters<M>>,
+}
+impl<M: Instrumenter> Clone for Handler<M> {
```
Nagaprasadvr, line 49: good refactoring , i like the pipelines  idea

examples/stream/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .with(tracing_subscriber::fmt::layer())
+        .init();
+
+    let Opts { config } = Opts::parse();
+    let config = std::fs::read_to_string(config).expect("Error reading config file");
+    let config = toml::from_str(&config).expect("Error parsing config");
+
+    vixen::stream::Server::builder()
```
Nagaprasadvr, line 91: looks clean

Nagaprasadvr (commented):
overall looks clean good to go

crates/core/src/instruction.rs
```diff
@@ -0,0 +1,328 @@
...
+            },
+        });
+
+        let mut outer = instructions
+            .into_iter()
+            .map(|i| Self::parse_one(Arc::clone(&shared), i))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        if inner_instructions_none {
+            return Err(Missing::InnerInstructions.into());
```
Nagaprasadvr: Every Outer ixs may not have inner ixs, i guess we should just ignore here and return an empty vector wyt @kespinola

crates/core/src/instruction.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub recent_blockhash: Vec<u8>,
+    pub accounts: AccountKeys,
+}
+
+#[derive(Debug)]
+pub struct InstructionUpdate {
+    pub program: Pubkey,
+    pub accounts: Vec<Pubkey>,
+    pub data: Vec<u8>,
+    pub shared: Arc<InstructionShared>,
```
Nagaprasadvr, line 72: for parsers we only need program , accounts , data 
 are we using this InstructionShared data somewhere else?

Nagaprasadvr: we can have AccountsShared instead of InstructionShared ig , and pass accounts as a vec of indices wyt?

Nagaprasadvr (changes requested):

crates/core/src/instruction.rs
ray-kast: >  are we using this InstructionShared data somewhere else?

`InstructionShared` is there to allow for more complex use cases that might need to e.g. store the slot or signature in which an instruction occurred in a database.

> we can ... pass accounts as a vec of indices wyt?

Honestly the `Pubkey` type is 256-bit and impls `Copy` so it's not gonna be some huge RAM or memory bandwidth sink.  Shouldn't be too big of a deal to store decoded account keys and it saves users having to do that error handling themselves.

It's also worth noting that for any use case that's advanced beyond what's offered here, users can still add a transaction parser and make alterations to the instruction parsing process for their own needs.  For right now if the `parser` crate is able to use the API presented here then we should be all set.  If we get feedback that certain common use cases don't work with this API then we can re-evaluate this stuff.

crates/core/src/instruction.rs
ray-kast: That's my bad, I thought that was an implementation detail of the Protobuf encoding from yellowstone but it looks like it's an oddity with `solana-transaction-status`.  I will just ignore that field.

crates/core/src/instruction.rs
Nagaprasadvr: Makes sense then!

crates/core/src/instruction.rs
Nagaprasadvr: cool

Nagaprasadvr (changes requested):
Can we update the READMEs considering current architectural changes

Nagaprasadvr (changes requested):
for Prometheus metrics to work ig we need to spin up a server at 0.0.0.0:3030 so metrics are pushed which can be consumed by the prometheus backed

or else we can also spin up a webserver using docker and expose port 3030 for use,
wyt @ray-kast  @kespinola 

check
https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/runtime/src/buffer.rs#L62

PR #22, opened 2024-09-09
jup program parsers 

Nagaprasadvr:
- [ ]  Parse AMM Programs (ix and accounts) used by Jupiter


ray-kast (changes requested):
Blocking this on #28.

crates/parser/src/jup_programs/raydium/account_parser/parser.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_amm_config_account_parsing() {
+        let parser = RaydiumProgramAccParser;
+
+        let account = account_fixture!("A1BBtTYJd4i3xU8D6Tc2FzU6ZN4oXZWXKZnCxwbHXr8x");
+
+        println!("acc:{:?}", account.clone().account.unwrap().data.len());
```
kespinola:
```suggestion
```

Capture log stuff in assertions as I imagine you are expecting the same value each run.

crates/parser/src/jup_programs/orca/ix_parser/parser.rs
```diff
@@ -0,0 +1,140 @@
...
+        if let OrcaProgramIx::Swap(accounts, data) = parsed {
+            assert_eq!(
+                accounts.whirlpool.to_string(),
+                "ENYEHSyduTbFN1xoSEGaLu7c1F8AqKucdscMuV5Yypy2".to_string()
+            );
+            assert_eq!(data.a_to_b, false);
+        } else {
+            panic!("Invalid Instruction");
+        }
+    }
```
kespinola:
```suggestion
        let ix = instruction_fixture!("3WC8LGHHs3wYzWef1YmLsRS96G1s5BV4XJYhzvypgWp1uGG16SxepFCd7FhHaTieW66Yn9JFR4tUPA1HYArgFZaA", 0, 1);

        let parsed = run_ix_parse!(parser, &ix);
        if let OrcaProgramIx::Swap(accounts, data) = parsed {
            assert_eq!(
                accounts.whirlpool.to_string(),
                "ENYEHSyduTbFN1xoSEGaLu7c1F8AqKucdscMuV5Yypy2".to_string()
            );
            assert_eq!(data.a_to_b, false);
        } else {
            panic!("Invalid Instruction");
        }
    }
```

What do you think of this dx? Caller provides the signature and path to the instruction they want. This example is the second inner instruction of the second instruction of the transaction.

We then write just that instruction to the disk at path like `ix-{sig}-0-1`

Nagaprasadvr: Can be done , but these indexes need to be know before itself , so users need to search for these indexes beforehand

crates/parser/src/jup_programs/raydium/account_parser/parser.rs
Nagaprasadvr: removing this logging which was used for debugging

crates/parser/src/jup_programs/raydium/ix_parser/parser.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub struct RaydiumProgramIxParser;
+
+impl Parser for RaydiumProgramIxParser {
+    type Input = InstructionUpdate;
+    type Output = RaydiumProgramIx;
+
+    fn id(&self) -> Cow<str> {
+        "yellowstone_vixen_parser::token_program::RaydiumProgramIxParser".into()
```
kespinola:
```suggestion
impl Parser for radium::InstructionParser {
    type Input = InstructionUpdate;
    type Output = RaydiumProgramIx;

    fn id(&self) -> Cow<str> {
        "yellowstone_vixen_parser::token_program::RaydiumProgramIxParser".into()
```

I'm been thinking about mod structure for the parser and think we should expose all pub on the root module in this case `radium` and then name have consistent parser names so `InstructionParser` and `AccountParser`

kespinola: Is this ID correct? token_program?

`yellowstone_vixen_parser::jup::radium::InstructionParser`

kespinola (changes requested):
Lets go over the package structure for parsers.

crates/parser/src/jup_programs/raydium/ix_parser/parser.rs
Nagaprasadvr: will be doing this in a new pr

crates/parser/src/token_extension_program/ix_parser/extensions/confidential_transfer_fee.rs
```diff
@@ -159,7 +159,7 @@ mod proto_parser {
         ConfidentialWithdrawWithheldTokensFromMintAccounts, DisableHarvestToMintAccounts,
         EnableHarvestToMintAccounts, InitializeConfidentialTransferFeeConfigAccounts,
     };
-    use crate::helpers::{FromVecPubkeyToVecString, IntoProto};
+    use crate::helpers::{proto_helpers::FromVecPubkeyToVecString, IntoProto};
```
kespinola: If this is already in helpers crate then call this module `proto` `helpers::proto::FromVecPubkeyToVecString`

kespinola (changes requested):

crates/parser/src/raydium/account_parser.rs
```diff
@@ -0,0 +1,118 @@
...
+impl ProgramParser for AccountParser {
+    #[inline]
+    fn program_id(&self) -> yellowstone_vixen_core::Pubkey {
+        RADIUM_V3_PROGRAM_ID.to_bytes().into()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use yellowstone_vixen_mock::{account_fixture, run_account_parse, FixtureData};
```
kespinola, line 94:
```suggestion
    use yellowstone_vixen_mock::account_fixture;
```

We just need account_fixture now.

crates/parser/src/orca/instruction_parser.rs
```diff
@@ -0,0 +1,139 @@
...
+                ))
+            },
+            _ => Err(ParseError::from("Unknown instruction")),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use yellowstone_vixen_mock::{run_ix_parse, tx_fixture, FixtureData};
```
kespinola:
```suggestion
    use yellowstone_vixen_mock::tx_fixture;
```

Just need to expose tx_fixture now

kespinola (changes requested):
Just cleanup on the test imports and we are g2g.

kespinola (approved):

PR #24, opened 2024-09-11
ProtoBufs :parse  token program and token extensions accounts  

Nagaprasadvr:
- [ ] Add Token Program Protobufs
- [ ] Add Token Extensions Protobufs


crates/parser/src/helpers.rs
```diff
@@ -58,3 +59,20 @@ impl<A, D> ReadableInstruction<A, D> {
...
+
+pub trait IntoProtoData<O> {
+    fn into_proto_data(self) -> O;
+}
+
+pub fn pubkey_to_string(pubkey: SolanaPubkey) -> String { pubkey.to_string() }
+
+pub fn bytes_to_elgamal_pubkey(bytes: [u8; ELGAMAL_KEYPAIR_LEN / 2]) -> String {
+    bs58::encode(bytes).into_string()
+}
```
kespinola: Since L63, 74, 76 are type conversion represent them using the From trait.

kespinola (changes requested):

kespinola:
@Nagaprasadvr can you add a README.me to the streams examples about to listen using grpcurl.

crates/parser/src/helpers.rs
Nagaprasadvr: these types are from external crate so i guess we cant impl From Trait

ray-kast (approved):

crates/parser/src/helpers.rs
kespinola: SolanaPubkey(pubkey::Pubkey)
ElgamalPubkeyBytes([u8; ELGAMAL_KEYPAIR_LEN / 2])

and then define from traits on these structs.

Nagaprasadvr: fixed

kespinola (approved):

PR #28, opened 2024-09-27
Refactor parser crate

ray-kast:
- **Fix stream error with duplicate program IDs.**
- **Pass 1 at removing extraneous helpers from parser.**
- **Clean up error & pubkey types in the parser crate.**


kespinola (approved):
👍 Looks great. Removing the extra wrapper on instructions and additional traits help add readability and repeatability I was looking for for the parser.

Nagaprasadvr (approved):
gtg

PR #55, opened 2025-02-24
[Bugfix] - Fix build when adding `yellowstone-vixen` as a dependency through Github link

pileks:
When `yellowstone-vixen` is added as a crate/workspace dependency using its Github link, its `topograph` dependency fails to build, due to having a non-existent branch specified as the source.

The fix is to simply specify the exact revision the `0.4.0` version of `topograph` that was present in `Cargo.lock`.

An alternative solution would be to simply specify the released [`0.4.0` version on crates.io](https://crates.io/crates/topograph).

I aimed to make as little of a change as possible, but if you'd like to see this change, I'd be happy to test it out.


crates/runtime/Cargo.toml
```diff
@@ -24,7 +24,7 @@ thiserror = "1.0.64"
 tokio = { version = "1.37.0", features = ["rt-multi-thread", "signal"] }
 topograph = { version = "0.4.0", features = [
     "tokio",
-], git = "https://github.com/ray-kast/topograph", branch = "v0.4-wip" }
+], git = "https://github.com/ray-kast/topograph", rev = "c91c01c25f605c4bc4461e9b9cbf95d8e4038972" }
```
kespinola: topograph is now published in crates.io. 

Do you mind updating to that?

kespinola (changes requested):

crates/runtime/Cargo.toml
pileks: Done ser 🫡

kespinola (approved):

kespinola:
@pileks thank you for the contribution. If you DM on X we have developer TG group that I'd like to add you to.

https://x.com/espiKnwldgBmbs

PR #63, opened 2025-03-21
`wip` Add meteora support

fernandodeluret:


codama_parser_generators/meteora.cjs
```diff
@@ -0,0 +1,48 @@
...
+  node,
+  renderVixenVisitor(
+    path.join(rootDir, "crates/parser/src", project, "generated"),
+    {
+      sdkName,
+      crateFolder: path.join(rootDir, "crates/parser/src", project),
+      formatCode: true,
+    }
+  )
+);
```
kespinola: When I was running generator for yellowstone-shield I didn't get any conflict from having the parser and client in the same crate. For these parsers I'd like to try that so we have

```
/crates
  /parsers
    /meteora
      src/
        generated/*
      Cargo.toml
idls/
  meteroa.json
  pumpfun.json
package.json
codama.mjs
```

We have a single `codama.mjs` script that loops through all IDLs in the `idls` folder. You run through `npm run generate`

I think the sdkName in this case is either `crate` or `super`. I hope it works like that or its just the same name as the parser.

The name of each parser cargo package should be `yellowstone-vixen-{protocol}-parser`

kespinola (changes requested):

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
-members = ["crates/*", "examples/*"]
-exclude = ["examples/streams-ts/"]
+members = ["crates/*", "examples/*", "crates/parsers_codama/yellowstone-vixen-meteora-parser", "crates/parsers_codama/yellowstone-vixen-pump-fun-parser"]
```
kespinola: Just have this be flat then with `crates/meteora-parser` but have the crate name be `yellowstone-vixen-pumpfun-parser`

crates/parsers_codama/yellowstone-vixen-meteora-parser/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+solana-sdk = { version = "2.0.3", optional = true }
+thiserror = "1.0.64"
+yellowstone-vixen-core = { workspace = true }
+
+[features]
+anchor = ["dep:anchor-lang"]
+anchor-idl-build = ["anchor", "anchor-lang?/idl-build"]
+serde = ["dep:serde", "dep:serde_with"]
+test-sbf = []
+fetch = ["dep:solana-client", "dep:solana-sdk"]
```
kespinola: This needs some fetchers for `proto` I believe.

crates/parsers_codama/yellowstone-vixen-meteora-parser/src/generated_parser/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! This code was AUTOGENERATED using the codama library.
+//! Please DO NOT EDIT THIS FILE, instead use visitors
+//! to add features, then rerun codama to update it.
+//!
+//! <https://github.com/codama-idl/codama>
+//!
+
+pub mod accounts_parser;
+
+pub mod instructions_parser;
```
kespinola: Ok both the rust sdk and vixen want to write to the generated file.

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "yellowstone-vixen",
```
kespinola, line 2:
```suggestion
  "name": "yellowstone-vixen",
  "private": true
```

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "codama_parser_generators/meteora.cjs",
```
kespinola:
```suggestion
```

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@codama/nodes": "^1.2.10",
+    "@codama/nodes-from-anchor": "^1.1.10",
+    "@codama/renderers-core": "^1.0.12",
+    "@codama/renderers-rust": "^1.0.18",
+    "@codama/renderers-vixen-parser": "^1.0.10",
+    "@codama/visitors-core": "^1.2.10"
+  }
```
kespinola: These are devDependencies.

kespinola (changes requested):
Almost there. Generate script is good just feedback on folder structure.

crates/parsers_codama/yellowstone-vixen-meteora-parser/Cargo.toml
fernandodeluret: Are you ok if we only include all the parsers added in this PR without taking into account the Vixen streams changes, and then we can do another PR for Vixen stream related changes?

codama_parser_generators/meteora.cjs
kespinola: Switching to packaging idl and codama.mjs with each parser crate.

PR #87, opened 2025-06-01
Stake Pool Program

aoikurokawa:


crates/parser/src/stake_pool_program/account_parser.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        if let Ok(validator_list) = try_from_slice_unchecked::<ValidatorList>(data_bytes) {
+            return Ok(SplStakePoolProgramState::ValidatorList(validator_list));
+        }
+
+        Err(yellowstone_vixen_core::ParseError::from(
+            "Invalid Account".to_owned(),
+        ))
+    }
```
fernandodeluret: I think it might cause trouble to deserialize with `try_from_slice_unchecked` because if the buffer is bigger it's always going to match the 1st if let branch. Also if you could update this to use a match statement like for example here https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/parser/src/token_program/account_parser.rs#L19-L30 would be better to check exhaustiveness.

I'm not sure what strategy the program uses for accounts (might be discriminators or account lenght), but in general you need to deserialize based on that

fernandodeluret: It looks like they use the 1st byte for account type https://github.com/solana-program/stake-pool/blob/main/program/src/state.rs but they also have some helper methods to check, so probably that would be the best to use (also in some cases they implement Pack, so there I think using unpack() would be the way to go)

crates/parser/src/stake_pool_program/instruction_parser.rs
```diff
@@ -0,0 +1,729 @@
...
+    fn program_id(&self) -> yellowstone_vixen_core::Pubkey {
+        spl_stake_pool::id().to_bytes().into()
+    }
+}
+
+impl InstructionParser {
+    pub(crate) fn parse_impl(
+        ix: &yellowstone_vixen_core::instruction::InstructionUpdate,
+    ) -> yellowstone_vixen_core::ParseResult<StakePoolProgramIx> {
+        let ix_type = StakePoolInstruction::deserialize(&mut ix.data.as_slice()).unwrap();
```
fernandodeluret: Similar comment here, we should be using `try_from_slice` as they are in the program (https://github.com/solana-program/stake-pool/blob/main/program/src/processor.rs#L3440), this adds the buffer lenght check on top of deserialize. Also if you can please handle the error instead of using unwrap this function should return a Result and just use `?` here

```suggestion
        let ix_type = StakePoolInstruction::try_from_slice(ix.data.as_slice()).unwrap();
```

crates/parser/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod block_meta;
 
 #[cfg(feature = "raydium")]
 pub mod raydium;
+pub mod stake_pool_program;
```
fernandodeluret: Nitpicky, but probably would be good to have this also under a feature flag as we have with the other program modules if you could :pray:

fernandodeluret: Actually it would be nicer if we had this in its own crate outside the parser crate (we plan to do the same and move out token and token22 too). Do you think you would be able to move your logic to its own containing crate?

fernandodeluret (changes requested):

crates/parser/src/stake_pool_program/instruction_parser.rs
aoikurokawa: Thanks, addresses [3080f03](https://github.com/rpcpool/yellowstone-vixen/pull/87/commits/3080f034b0d3f2daf3a9e0e61edf8e77181e1b1d)

crates/parser/src/stake_pool_program/account_parser.rs
aoikurokawa: Thank you for your advise!
I changed the way to deserialize.

```rust
    pub fn try_unpack(data_bytes: &[u8]) -> yellowstone_vixen_core::ParseResult<Self> {
        let first_byte = data_bytes[0];

        match first_byte {
            0 => Err(yellowstone_vixen_core::ParseError::from(
                "Invalid Account".to_owned(),
            )),
            1 => {
                let stake_pool = try_from_slice_unchecked(data_bytes)?;
                return Ok(SplStakePoolProgramState::StakePool(stake_pool));
            },
            2 => {
                let validator_list = try_from_slice_unchecked(data_bytes)?;
                return Ok(SplStakePoolProgramState::ValidatorList(validator_list));
            },
            _ => Err(yellowstone_vixen_core::ParseError::from(
                "Invalid Account".to_owned(),
            )),
        }
    }

```

crates/parser/src/lib.rs
aoikurokawa: I moved it to its own crate! Thank you!

aoikurokawa:
@fernandodeluret 
I really appreciate your review!!

crates/stake-pool-parser/src/account_parser.rs
```diff
@@ -0,0 +1,71 @@
+use spl_pod::solana_program::{self, borsh1::try_from_slice_unchecked};
+use spl_stake_pool::state::{StakePool, ValidatorList, ValidatorStakeInfo};
+
+/// SplStakePool Program State
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug)]
+pub enum SplStakePoolProgramState {
+    StakePool(StakePool),
+    ValidatorStakeInfo(ValidatorStakeInfo),
```
fernandodeluret: I think we can remove this here, because `ValidatorStakeInfo` it's not an account, it's each of the elements of the `validators` Vec in `ValidatorList`

crates/stake-pool-parser/src/account_parser.rs
```diff
@@ -0,0 +1,71 @@
...
+    ValidatorList(ValidatorList),
+}
+
+impl SplStakePoolProgramState {
+    pub fn try_unpack(data_bytes: &[u8]) -> yellowstone_vixen_core::ParseResult<Self> {
+        let first_byte = data_bytes[0];
+
+        match first_byte {
+            0 => Err(yellowstone_vixen_core::ParseError::from(
+                "Invalid Account".to_owned(),
```
fernandodeluret: Nitpicky, but maybe I would return a different error here, like "Uninitialized Account", because that case is specifically handled in their accountsType

fernandodeluret (commented):
@aoikurokawa Thanks for all the contributions! And apologies for the very delayed response :pray: 

I have tested this, and Instructions are looking great, you are even correctly handling the optional accounts and all parsing results seems correct, so that's great!

On the other hand, I think accounts might be deserializing incorrectly and silently going wrong. Correct me if I'm wrong, but you are still using `try_from_slice_unchecked` which does not check the bytes buffer length and if I run this and test if there is remaining data there, there are bytes not being deserialized. 

Maybe you have been using this parser more than me and there is something I'm missing, but in their program they also use `try_from_slice_unchecked` in some of the instructions, but because they still do some more things with the account_info data besides that (and that function does not update the pointer in the buffer contrary to `deserialize()` also from Borsh) so they handle the data independently and write to the account again at the end of ixs. My concern is that we don't have that account_info around any more, any bytes we don't deserialize we just discard them (for validators in particular idk if we shouldn't be using the `max_validators` len in the header rather than the Vec len to deserialize). If you have any additional context regarding this please let me know.

aoikurokawa:
@fernandodeluret 
Thanks for the thorough review and testing! I appreciate you taking the time to verify the instruction parsing works correctly.

Regarding your concern about `try_from_slice_unchecked` potentially missing data in account deserialization - I've done some additional investigation and testing:

1. **Following official patterns**: My implementation uses the exact same approach as the official SPL stake pool program's RPC client functions ([`get_stake_pool`](https://github.com/solana-program/stake-pool/blob/f6edd815fee562f5758ea53c5e3c027b12d3dd47/clients/cli/src/client.rs#L24) and [`get_validator_list`](https://github.com/solana-program/stake-pool/blob/f6edd815fee562f5758ea53c5e3c027b12d3dd47/clients/cli/src/client.rs#L34)). They also use `try_from_slice_unchecked` for deserializing these account types.

2. **Real-world testing**: I've tested this with live devnet data and the parsing works correctly, producing complete and accurate account structures. For example, I successfully parsed a complete `StakePool` account with all nested data intact.
https://github.com/jito-foundation/jito-bell/blob/ak/stake_pool/examples/stake_pool/src/main.rs

```txt
2025-06-28T02:23:24.506286Z  INFO stake_pool: value=StakePool(StakePool { account_type: StakePool, manager: aaaDerwdMyzNkoX1aSoTi3UtFe2W45vh5wCgQNhsjF8, staker: aaaDerwdMyzNkoX1aSoTi3UtFe2W45vh5wCgQNhsjF8, stake_deposit_authority: 74opVa3v51hUmTrsZn8YusZw4fXB16vGQY4WYHt9UegR, stake_withdraw_bump_seed: 253, validator_list: nZ5vUrsJjHcvkJsfKP1b1RgSEZUMJmwFpk7NksTeX5A, reserve_stake: rrWBQqRqBXYZw3CmPCCcjFxQ2Ds4JFJd7oRQJ997dhz, pool_mint: J1toso1uCk3RLmjorhTtrVwY9HJ7X8V9yYac6Y7kGCPn, manager_fee_account: DH7tmjoQ5zjqcgfYJU22JqmXhP5EY1tkbYpgVWUS2oNo, token_program_id: TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA, total_lamports: 280351583865, pool_token_supply: 279820515100, last_update_epoch: 904, lockup: Lockup { unix_timestamp: 0, epoch: 0, custodian: 11111111111111111111111111111111 }, epoch_fee: Fee { denominator: 100, numerator: 4 }, next_epoch_fee: None, preferred_deposit_validator_vote_address: None, preferred_withdraw_validator_vote_address: None, stake_deposit_fee: Fee { denominator: 0, numerator: 0 }, stake_withdrawal_fee: Fee { denominator: 1000, numerator: 1 }, next_stake_withdrawal_fee: None, stake_referral_fee: 0, sol_deposit_authority: None, sol_deposit_fee: Fee { denominator: 0, numerator: 0 }, sol_referral_fee: 0, sol_withdraw_authority: None, sol_withdrawal_fee: Fee { denominator: 1000, numerator: 1 }, next_sol_withdrawal_fee: None, last_epoch_pool_token_supply: 277824303692, last_epoch_total_lamports: 278351583865 })
2025-06-28T02:23:24.506452Z  INFO stake_pool: value=DepositSol(DepositSolAccounts { stake_pool: Jito4APyf642JPZPx3hGc6WWJ8zPKtRbRs4P815Awbb, stake_pool_withdraw_authority: 6iQKfEyhr3bZMotVkW6beNZz5CPAkiwvgV2CTje9pVSS, reserve_stake_account: rrWBQqRqBXYZw3CmPCCcjFxQ2Ds4JFJd7oRQJ997dhz, lamports_from: 2V6Abua9BY6Ga8HUeLWSLXh4Gm6oKsn3GpTzP4eYMFqT, pool_tokens_to: 22Mjmaea25LDrpEQyJonfV6ybkrDcxGDsoCqUH39Cw9m, manager_fee_account: DH7tmjoQ5zjqcgfYJU22JqmXhP5EY1tkbYpgVWUS2oNo, referrer_pool_tokens_account: 22Mjmaea25LDrpEQyJonfV6ybkrDcxGDsoCqUH39Cw9m, pool_mint: J1toso1uCk3RLmjorhTtrVwY9HJ7X8V9yYac6Y7kGCPn, system_program: 11111111111111111111111111111111, token_program: TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA, deposit_authority: None }, DepositSolData { arg: 2000000000 })
```

I'm very welcome to more review and happy to investigate further or change the approach if needed.

Nagaprasadvr:
@aoikurokawa can u fix clippy just run
 ```cargo clippy``` and update the code accordingly

aoikurokawa:
@Nagaprasadvr 
Thank you for reviewing!
I fixed the clippy issue now!

Nagaprasadvr:
@aoikurokawa can u add a description of all the changes u made i see a lot of diffs

fernandodeluret:
@aoikurokawa If we can make CI to pass I'll merge it before we merge other additional changes

crates/parser/src/lib.rs
```diff
@@ -18,6 +18,8 @@ mod helpers;
 #[cfg(feature = "block-meta")]
 pub mod block_meta;
 
+#[cfg(feature = "raydium")]
+pub mod raydium;
```
fernandodeluret: I think this is the change causing the CI lint

aoikurokawa: @fernandodeluret 
Thank you!
How about now?

fernandodeluret: I'm checking what it's complaining about now

aoikurokawa: Seems clippy issues are coming from what i have not changed.

fernandodeluret: True, give me a couple of mins. I'm running CI on my own PR (locally clippy worked fine for me)

fernandodeluret: Just passed https://github.com/rpcpool/yellowstone-vixen/pull/111. May be your cargo lock file. Could you try to run and let's see what that says for you locally:
```
cargo build
cargo clippy --all-targets --tests -- -Dwarnings

aoikurokawa: Thanks!
it works on my local now! Let's see passing the CI.

aoikurokawa:
Thank you for your support!! @fernandodeluret

PR #91, opened 2025-06-05
Bump yellowstone and expose from_slot filter

fernandodeluret:


examples/prometheus/src/main.rs
```diff
@@ -58,6 +58,7 @@ fn main() {
         .block_meta(Pipeline::new(BlockMetaParser, [Logger]))
         .metrics(vixen::metrics::Prometheus)
         .commitment_level(yellowstone_vixen::CommitmentLevel::Confirmed)
+        // .from_slot(slot_number)
```
sonicfromnewyoke, line 61: shouldn't it be in this way?
```suggestion
         .from_slot(None)
```

fernandodeluret: Hey Sonic! Thanks for taking the time to review! No, in this particular one we are receiving the `u64` instead of the `Option<u64>` (because in general you would only call the method if you want to actually set the filter)

sonicfromnewyoke: oh, got it. THanks

kespinola (approved):

Issue #90, opened 2025-06-04
🚀 RFC: Support for Multiple Data Sources


fernandodeluret:
## Problem Statement

Currently, Yellowstone Vixen is tightly coupled to Dragon's Mouth gRPC streams as the sole data source. While this provides excellent real-time capabilities, there are scenarios where developers might benefit from alternative data sources or hybrid approaches:

- **Historical data processing**: Backfilling or reprocessing historical events
- **Optionality**: Leverage other solutions like Dragon's Mouth or Fumarole for live updates

## Proposed Solution

Introduce a pluggable data source architecture that maintains the existing parser + handler framework while allowing different input mechanisms.

### Proposed API

Here's an initial API design that maintains the familiar builder pattern:

```rust
vixen::Runtime::builder()
    .source(Source::new(DragonsMouth))
    .source(Source::new(SolanaAccountsRpc))
    .account(Pipeline::new(TokenProgramParser, [Handler]))
    .build(config)
    .run();
```

This approach would allow:
- **Multiple concurrent sources**: Register different data sources that concurrently feed into the same pipeline
- **Source-agnostic parsers**: Existing parsers work with any data source
- **Familiar API**: Minimal changes to current usage patterns
- **Backfilling**: Concurrent historical account processing alongside real-time streams for complete data coverage
- **Shared filters**: Use the defined filters for all relevant sources

### Potential Data Sources

**Immediate candidates:**
- 🔥 **Dragon's Mouth gRPC** (current, remains default)
- 🏗️ **RPC Account Backfilling**: Concurrent historical account processing alongside real-time streams for complete data coverage

**Future possibilities:**
- **Yellowstone Fumarole**: Upcoming geyser implementation with reliability and data persistence features
- **Multi-source Aggregation**: Combine multiple sources for redundancy
- **Custom Data Sources**: Plugin system for community-contributed sources

## Benefits

- **Flexibility**: Choose the right data source for your use case
- **Development Experience**: Easier local development and testing
- **Reliability**: Fallback options for production systems
- **Backwards Compatibility**: Existing Dragon's Mouth implementations remain unchanged

### Architecture Changes
- Introduce a `Source` trait
- Make the Vixen runtime support a vector of sources to process concurrently
- Maintain current `Pipeline` and parser architecture
- Maintain current `config` format

### Challenges to Discuss
- **Developer experience**: Best approach for balance flexibility and ease of use
- **Data consistency**: Ensuring consistent behavior across sources
[...]

PR #96, opened 2025-06-12
Multiple Sources

kespinola:
## Changes
- Support multiple sources through registration via `Source` trait

```
    vixen::Runtime::builder()
        .source(YellowstoneGrpcSource)
        .source(SolanaRpcSource)
        .account(Pipeline::new(TokenProgramAccParser, [Logger]))
        .account(Pipeline::new(TokenExtensionProgramAccParser, [Logger]))
        .instruction(Pipeline::new(TokenExtensionProgramIxParser, [Logger]))
```

#90

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -201,7 +201,7 @@ impl<M: MetricsFactory> Runtime<M> {
     pub async fn try_run_async(self) -> Result<(), Error> {
         enum StopType<S, X> {
             Signal(S),
-            Buffer(Result<std::convert::Infallible, Error>),
+            Buffer(Result<(), Error>),
             Exporter(Result<Result<stop::StopCode, X>, tokio::task::JoinError>),
         }
 
@@ -268,7 +268,11 @@ impl<M: MetricsFactory> Runtime<M> {
 
         let stop_ty = tokio::select! {
             s = signal => StopType::Signal(s),
-            b = buffer.wait_for_stop() => StopType::Buffer(b),
+            b = buffer.wait_for_stop() => {
+                println!("stop type buffer");
+
+                StopType::Buffer(b)
+            },
             Some(x) = &mut exporter => StopType::Exporter(x),
         };
 
@@ -285,9 +289,8 @@ impl<M: MetricsFactory> Runtime<M> {
                 "Signal handler returned None",
             )
             .into()),
-            // Not sure why the compiler couldn't figure this one out
-            StopType::Buffer(Ok(o)) => match o {},
-            StopType::Signal(Err(e)) | StopType::Buffer(Err(e)) => Err(e),
+            StopType::Buffer(result) => result,
+            StopType::Signal(Err(e)) => Err(e),
             StopType::Exporter(Ok(Ok(..))) => {
                 Err(Error::MetricsExporter("Exporter stopped early".into()))
             },
--- a/crates/runtime/src/buffer.rs
+++ b/crates/runtime/src/buffer.rs
@@ -30,12 +30,11 @@ impl Buffer {
             .and_then(std::convert::identity)
     }
 
-    // TODO: use never
-    pub async fn wait_for_stop(&mut self) -> Result<std::convert::Infallible, crate::Error> {
+    pub async fn wait_for_stop(&mut self) -> Result<(), crate::Error> {
         (&mut self.0)
             .await
             .map_err(|e| std::io::Error::from(e).into())
-            .and_then(|r| r.and(Err(crate::Error::ClientHangup)))
+            .map(|_| ())
     }
 }
 
@@ -173,7 +172,10 @@ impl Buffer {
 
                         let update = match event {
                             Event::Update(Some(u)) => u,
-                            Event::Update(None) => break Err(crate::Error::ServerHangup),
+                            Event::Update(None) => {
+                                tracing::warn!("Server stopped sending updates");
+                                break Ok(StopCode::default());
+                            },
                             Event::Stop(c) => break Ok(c),
                         };
 
--- a/crates/runtime/src/util.rs
+++ b/crates/runtime/src/util.rs
@@ -142,7 +142,7 @@ pub mod stop {
     use tokio::sync::oneshot;
 
     /// A signal that a task has received a request to stop.
-    #[derive(Debug)]
+    #[derive(Debug, Default)]
     #[must_use = "Consider calling .as_unit()"]
     #[allow(missing_copy_implementations)]
     #[repr(transparent)]


crates/runtime/src/sources/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+use vixen_core::Filters;
+use yellowstone_grpc_proto::{geyser::SubscribeUpdate, tonic::Status};
+
+use crate::config::YellowstoneConfig;
+
+/// A `Source` implementation for the Yellowstone gRPC API.
+pub mod yellowstone_grpc;
+
+/// A `Source` implementation for Solana Accounts backfilling through the RPC API.
+pub mod solana_accounts_rpc;
```
kespinola: How much does the dependency tree change if we split out sources into its own crate or a crate per source. There is value if vixen runtime doesn't share many crates with sources.

Ideally runtime has no dependencies on any sort of specific source. And each source crate doesn't share crates with each other except for `yellowstone-vixen-sources`

crates/runtime/src/sources/solana_accounts_rpc.rs
```diff
@@ -0,0 +1,157 @@
...
+            self.get_commitment_config(),
+        );
+
+        for (filter_id, prefilter) in filters.parsers_filters {
+            if let Some(account_prefilter) = prefilter.account {
+                for program in account_prefilter.owners {
+                    let program_id = Pubkey::new_from_array(program.0);
+
+                    // TODO: Only get pubkey from get_program_accounts_with_config() using data_slice
+                    //  and then use get_multiple_accounts() for ~batching (process batches concurrently)
```
kespinola: Is this to get less data coming over the wire at a time? I've generally observed the key to getProgramAccounts is to not ever use a filter don't have the rpc server have to perform filter logic.

examples/prometheus/src/main.rs
```diff
@@ -51,6 +53,7 @@ fn main() {
     let config = toml::from_str(&config).expect("Error parsing config");
 
     vixen::Runtime::builder()
+        .source(YellowstoneGrpcSource::new())
```
kespinola, line 51:
```suggestion
        .source(YellowstoneGrpcSource)
```

If we don't need wrapper struct `Source::new(YellowstonGrpcSource)` then lets enforce need to expose new and runtime calls it.

crates/runtime/src/lib.rs
```diff
@@ -274,4 +290,27 @@ impl<M: MetricsFactory> Runtime<M> {
...
+        let mut set = JoinSet::new();
+
+        while let Some(mut source) = self.sources.pop() {
+            let tx = tx.clone();
+            source.config(self.yellowstone_cfg.clone());
+            source.filters(filters.clone());
+
+            set.spawn(async move {
+                source.connect(tx).await.expect("Source connection failed");
+            });
```
kespinola: A note that we don't gracefully shutdown sources. They don't receive a shutdown signal from the parent to do cleanup. In practice I don't think a problem for the sources we have now but we don't provide a means for sources to handle graceful shutdown.

crates/runtime/src/sources/solana_accounts_rpc.rs
```diff
@@ -0,0 +1,157 @@
...
+                    GlobalFilters {
+                        commitment: Some(CommitmentLevel::Processed),
+                    },
+                ..
+            }) => CommitmentConfig::processed(),
+            _ => CommitmentConfig::confirmed(),
+        }
+    }
+}
+
```
kespinola: This seems better suited to From trait `From<Filters> for CommitmentConfig`

crates/runtime/src/sources/solana_accounts_rpc.rs
```diff
@@ -0,0 +1,157 @@
...
+                                    pubkey: acc_pubkey.as_array().to_vec(),
+                                    lamports: account.lamports,
+                                    owner: account.owner.as_array().to_vec(),
+                                    executable: account.executable,
+                                    rent_epoch: account.rent_epoch,
+                                    data: account.data,
+                                    write_version: 0,
+                                    txn_signature: None,
+                                }),
+                                slot: 0,
```
kespinola: Rpc has context which is the slot the accounts are pulled from. We need to set slot its very important for indexers.

crates/runtime/src/sources/solana_accounts_rpc.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    fn get_filters(&self) -> &Option<Filters> { &self.filters }
+
+    fn get_config(&self) -> Option<YellowstoneConfig> {
+        self.config.clone().map(
+            |SolanaAccountsRpcConfig { endpoint, timeout }| YellowstoneConfig {
+                endpoint,
+                timeout,
+                x_token: None,
+            },
```
kespinola:
```suggestion
Into::into
```

See can represent in from trait

crates/runtime/src/sources/yellowstone_grpc.rs
```diff
@@ -0,0 +1,69 @@
...
+        let (_sub_tx, stream) = client.subscribe_with_request(Some(filters.into())).await?;
+
+        let mut stream = std::pin::pin!(stream);
+
+        while let Some(update) = stream.next().await {
+            let res = tx.send(update).await;
+            if res.is_err() {
+                tracing::error!("Failed to send update to buffer");
+            }
+        }
```
kespinola: Can you add issue to handle ping-pong of the subscription? Examples in the dm or das repos.

kespinola (commented):
Looking good.

From what we have these are my thoughts on adjustments:

1. Concurrency per source and filter combo. For example, RPC + GRPC and 2 account parsers then running 4 request workers that all feed to the same processing pipeline. This is good for GRPC which benefits from spreading over several connections to aid in back pressure that may occur in the grpc trying to push packets over its window size.

crates/runtime/src/sources/mod.rs
fernandodeluret: Yes agreed! This is actually something I was thinking about (mainly to don't couple runtime against an specific version of solana client crates because RPC source), I'll have each source in its own crate :handshake:

crates/runtime/src/sources/solana_accounts_rpc.rs
fernandodeluret: Yes true, I only implemented it this way to avoid creating a wrapper for the struct (because both are defined outside the crate). Do you think would be better with the wrapper (tuple struct) and impl From?

crates/runtime/src/sources/solana_accounts_rpc.rs
fernandodeluret: You got me with this one, it's true that there is an option for asking for context but we don't get any in the response, seems they are deleting it here: (and I might be wrong but even if they weren't this is only returning a context for the whole request not per account)

https://github.com/anza-xyz/agave/blob/master/rpc-client-types/src/response.rs#L27-L32
```rust
impl<T> OptionalContext<T> {
    pub fn parse_value(self) -> T {
        match self {
            Self::Context(response) => response.value,
            Self::NoContext(value) => value,
        }
    }
}
```

Using the solana library the only way I saw to get slot info for the account is fetching each one of them individually with `get_account()` (even get_multiple_accounts returns only one slot value)

crates/runtime/src/lib.rs
fernandodeluret: Yes also true, the only thing we are doing is droping the handlers at the end of the runtime which aborts all `Source`s tasks, I was thinking to also expose a way for cleaning for Sources in the future, but if you want we can add it here

examples/prometheus/src/main.rs
fernandodeluret: Hmm, I think we could do something similar but not exactly that, because to pass it like that we need the struct to be a unit struct without data (if I'm not wrong), but what we could do is something like this:
 ```rust
  .source::<YellowstoneGrpcSource>()
  ``` 
 
 (we need to enforce `new()` through another trait though Source can't return owned values cause it's `dyn`, but I tried and works well)

crates/runtime/src/sources/solana_accounts_rpc.rs
fernandodeluret: I followed the trail a bit more on this, and I think I found the exact place where they are dropping the account slot for get_program_accounts (in the RPC I mean):

https://github.com/anza-xyz/agave/blob/master/accounts-db/src/accounts.rs#L341
```rust
    fn load_while_filtering<F: Fn(&AccountSharedData) -> bool>(
        collector: &mut Vec<TransactionAccount>,
        some_account_tuple: Option<(&Pubkey, AccountSharedData, Slot)>,
        filter: F,
    ) {
        if let Some(mapped_account_tuple) = some_account_tuple
            .filter(|(_, account, _)| account.is_loadable() && filter(account))
            .map(|(pubkey, account, _slot)| (*pubkey, account))
        {
            collector.push(mapped_account_tuple)
        }
    }
```

But I think they are doing the same for get_account https://github.com/anza-xyz/agave/blob/master/runtime/src/bank.rs#L5035:

```rust
    pub fn get_account(&self, pubkey: &Pubkey) -> Option<AccountSharedData> {
        self.get_account_modified_slot(pubkey)
            .map(|(acc, _slot)| acc)
    }
```

 (the slot we get from context is the current slot for the `Bank` struct, I might be wrong on this though)

examples/prometheus/src/main.rs
kespinola: I take this statement back. We want the source to be able to register external dependencies like redis or kafka client.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
-members = ["crates/*", "examples/*"]
-exclude = ["examples/streams-ts/"]
+members = ["crates/*", "examples/*", "crates/sources/solana-accounts-rpc-source", "crates/sources/yellowstone-grpc-source"]
+exclude = ["examples/streams-ts/", "crates/sources/"]
```
kespinola: I'm still of the mind of having the crates be single directory. They all have suffixes to denote type of crate. 

At larger scale filters and sources would have their own repositories.

crates/runtime/src/lib.rs
```diff
@@ -335,4 +349,32 @@ impl<M: MetricsFactory> Runtime<M> {
...
+
+        let filters = self
+            .pipelines
+            .filters()
+            .commitment(self.commitment_filter)
+            .from_slot(self.from_slot_filter);
+
+        let mut set = JoinSet::new();
+
+        while let Some(mut source) = self.sources.pop() {
```
kespinola:
```suggestion
    for mut source in self.sources.drain(..) {
```

crates/sources/solana-accounts-rpc-source/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+                                        lamports: account.lamports,
+                                        owner: account.owner.as_array().to_vec(),
+                                        executable: account.executable,
+                                        rent_epoch: account.rent_epoch,
+                                        data: account.data,
+                                        write_version: 0,
+                                        txn_signature: None,
+                                    }),
+                                    slot: 0,
+                                    is_startup: true,
```
kespinola: 1. Since we can't get slot can you open an issue into anza repo and do some research to propose a solution?
2. Until then I think slot is important enough that we get_slot before making the get_program_accounts its not ideal because the get_slot can be served from different RPC node than the one serving the get_program but it will have to do for now.

crates/runtime/src/lib.rs
```diff
@@ -335,4 +349,32 @@ impl<M: MetricsFactory> Runtime<M> {
             }
         }
     }
+
+    /// Connect to the sources and return the runtime, the receiver for the updates, and the join set for the sources.
+    fn connect_to_sources(
+        mut self,
+    ) -> (Self, Receiver<Result<SubscribeUpdate, Status>>, JoinSet<()>) {
+        let (tx, rx) =
+            mpsc::channel::<Result<SubscribeUpdate, Status>>(self.buffer_cfg.sources_channel_size);
```
kespinola, line 363: Since we are mixing long running and one-shot sources I don't think its a good idea to add a buffer.

Since we drop this we can remove the join_set return from the sources and have them operate more independently. For example, use `join_all(get_program_account_calls).await` and yellowstone grpc simple while await like it had before this refactor.

```suggestion
```

crates/runtime/src/lib.rs
```diff
@@ -335,4 +349,32 @@ impl<M: MetricsFactory> Runtime<M> {
...
+        let mut set = JoinSet::new();
+
+        while let Some(mut source) = self.sources.pop() {
+            let tx = tx.clone();
+            source.config(self.yellowstone_cfg.clone());
+            source.filters(filters.clone());
+
+            set.spawn(async move {
+                source.connect(tx).await.expect("Source connection failed");
+            });
```
kespinola, line 382: Can you please add a static name method on the source and print that in this expect so the operator knows which source failed.

kespinola: This join_set is good because we need to exit out of vixen when all the connections close. If dev uses grpc will only exit when forced like we have now but rpc will complete which means we should exit out and close the program.

fernandodeluret: Regarding this one, I implemented it but the method is not static (I mean it's a normal method receiving `&self` instead of an associated fn) because of dyn traits restrictions. The other way is to have a different trait for the name and require that one (initially I didn't do it that way because it's a bit more boilerplate to someone creating Sources impl, but they are almost the same, so if you prefer I can use this 2nd one)

kespinola (commented):
LG2M

PR #98, opened 2025-06-19
Update RaydiumAmmV4 parser to handle optional accounts

fernandodeluret:


crates/raydium-amm-v4-parser/src/generated_sdk/instructions/swap_base_out.rs
```diff
@@ -78,10 +78,12 @@ impl SwapBaseOut {
...
-        accounts.push(solana_program::instruction::AccountMeta::new(
-            self.amm_target_orders,
-            false,
-        ));
+        if let Some(amm_target_orders) = self.amm_target_orders {
+            accounts.push(solana_program::instruction::AccountMeta::new(
+                amm_target_orders,
+                false,
+            ));
+        }
```
kespinola, line 86: The accounts list always has to be the same length shouldn't this add the program address as the account when its none?

crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
```diff
@@ -492,6 +533,32 @@ pub fn check_min_accounts_req(
...
+    expected_accounts_len: &mut usize,
+) -> Result<Option<solana_program::pubkey::Pubkey>, yellowstone_vixen_core::ParseError> {
+    if actual_accounts_len == *expected_accounts_len + 1 {
+        *expected_accounts_len += 1;
+        Ok(Some(next_account(accounts)?))
+    } else {
+        Ok(None)
+    }
+}
+
```
kespinola, line 569: This is nice. Can leave in the parser for now but also a good utility to export from vixen for those not using codama.

crates/raydium-amm-v4-parser/src/generated_sdk/instructions/swap_base_out.rs
fernandodeluret: No, actually they(Raydium) use a different approach for optional accounts, the accounts list is variable length, and you only pass the optional account if you want it, if not you don't pass anything. 

Codama already supports this through the optionalAccountStrategy = 'programId' | Omitted

crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
fernandodeluret: Yes.. it's a tricky function though.. that's why I kept it here. Because if a program uses this approach of not passing an account if it's optional and let's say they have some logic like this:

```rust
accounts_list = {
  acc1: Option<_>
  acc2: Option<_>
}

pub fn ix1(args: bool, accounts_list) {
  if args {
     acc1.unwrap()
  } else {
     acc2.unwrap()    
  }
}
```

This function would be wrong, because it's asumming that to have acc2 you have acc1 present (In general it's what they do though, like here https://github.com/raydium-io/raydium-amm/blob/master/program/src/processor.rs#L5442-L5451)(in general this is more of a limitation of omitting optional accounts in general)

kespinola (approved):
Same request to squash merge and write 1-2 line commit for release notes. Thank you.

PR #99, opened 2025-06-20
Expose tx_sig & slot to raydium handler

fernandodeluret:


crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
```diff
@@ -56,9 +56,16 @@ pub enum RaydiumAmmProgramIx {
 #[derive(Debug, Copy, Clone)]
 pub struct InstructionParser;
 
+#[derive(Debug)]
+pub struct InstructionUpdateOutput {
+    pub parsed_ix: RaydiumAmmProgramIx,
+    pub tx_signature: Vec<u8>,
+    pub slot: u64,
+}
+
```
kespinola: This should be a generic

```suggestion
pub struct InstructionOutput<I> {
    pub instruction: I,
    pub signature: Vec<u8>,
    pub slot: u64,
}

```

kespinola: Likely a from trait too for `ix.map(Into::into)`

kespinola (changes requested):

crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
fernandodeluret: But making it generic might not be necesary, because this is a per parser struct, so we only ever use it with its particular program. The generic is already set in the `handle` fn from the `Parser` trait and set through the `type Output =`

Here it's only implemented for 1 parser, but this is meant to do this same change to each parser (which I think it's the better so that we don't constraint in any way the Parser trait return, and give each parser freedom to do what they like)

Also regarding the From trait we could implement it but only to go from InstructionUpdateOutput -> Ixs (because the other way around we need data that it's not in the struct, I mean we could still set those fields to default values, but idk if that makes too much sense)

fernandodeluret: @kespinola maybe I can update this PR to expose the full `shared_data` with any commets you have for it

kespinola: Yes expose all shared_data but should have standard shape so parser crate has 

```rust
pub struct InstructionOutput<I> {
	    pub instruction: I,
	    pub shared_data: TransactionSharedData,
	}
```

crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
```diff
@@ -475,7 +484,11 @@ impl InstructionParser {
             },
         }
 
-        ix
+        ix.map(|ix| InstructionUpdateOutput {
+            parsed_ix: ix,
+            tx_signature,
+            slot,
+        })
     }
```
kespinola:
```suggestion
        ix.map(|ix| InstructionUpdateOutput::new(ix, shared))
        
    }
```

kespinola (approved):
This looks good. Why are we adding the `Box<Error>` just so have info for the release notes.

When you merge can you squash and write description to be presented in the changelog github pulls from the commit messages.

fernandodeluret:
> This looks good. Why are we adding the `Box<Error>` just so have info for the release notes.
> 
> When you merge can you squash and write description to be presented in the changelog github pulls from the commit messages.

The box error is because in the new version of rust (I think it's  1.88) We were getting those lints errors, so it was just a minor fix for that

PR #104, opened 2025-07-09
`feat` Expose transaction included and required accounts using FilterPipeline

fernandodeluret:


crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
```diff
@@ -526,6 +547,45 @@ impl InstructionParser {
...
+    ///
+    /// **The Program ID of the Parser program will always be included in this list
+    pub fn required_accounts(
+        mut self,
+        required_accounts: impl IntoIterator<Item = yellowstone_vixen_core::Pubkey>,
+    ) -> Self {
+        self.required_accounts.extend(required_accounts);
+
+        self
+    }
```
kespinola: Can these be exposed through vixen crate so not repeated? Can store in a trait that fully defines the methods and then have the cfg impl it for the InstructionParser

kespinola (changes requested):
Nice API just want to talk about where to put the code for the extra filtering. Its easier to keep parsers aligned based on crate version rather than generating code so like to expose more from vixen crates and not generate directly in the parser.

crates/raydium-amm-v4-parser/src/generated_parser/instructions_parser.rs
fernandodeluret: I think this 9db8cae506e6bbcf478405df65030a8702cb77e1 did the trick. 

The API almost didn't change:
```rust
RaydiumAmmV4IxParser.filter().include_accounts([Pubkey::from_str("").unwrap()])
```

And now crates only conditionally compile this:
```rust
#[cfg(feature = "prefilters")]
impl CustomPrefilters for InstructionParser {}
```

fernandodeluret: I continued thinking and updated the API to also support direct strings and Parsers:
```rust
  RaydiumAmmV4IxParser.filter()
      .include_accounts(["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"])
      .required_accounts([RaydiumAmmV4AccParser]),
```

What do you think? @kespinola

examples/prometheus/src/main.rs
```diff
@@ -51,7 +51,12 @@ fn main() {
         .source(YellowstoneGrpcSource::new())
         .source(SolanaAccountsRpcSource::new())
         .account(Pipeline::new(RaydiumAmmV4AccParser, [Logger]))
-        .instruction(Pipeline::new(RaydiumAmmV4IxParser, [Logger]))
+        .instruction(Pipeline::new(
+            RaydiumAmmV4IxParser.filter()
+                .include_accounts(["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"])
+                .required_accounts([RaydiumAmmV4AccParser]),
+            [Logger],
```
kespinola:
```suggestion
        .instruction(Pipeline::new(
            Pipe::new()
              .filter(
                  Filter::build().
                    .include_accounts(
        ["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"]
                    )
                    .parser(RaydiumAmmV4AccParser),
            [Logger],
```

kespinola: Just sudo coding some alternate API for adding more filter options. In this version we have a constructor that takes the parser and extra filters and builds up a new parser out of the both so we aren't handling merge logic within the parser specification its done outside.

kespinola:
```rust
impl Preflight for RaydiumAmmV4AccParser {
  fn preflight() -> Preflight {
    Preflight::new().include_accounts([])
  }
}
```

```suggestion
        .instruction(Pipeline::new(RaydiumAmmV4AccParser, [Logger],)
```

And in this example you impl Preflight in the target binary but this means can't do dynamic configuration without macros.

examples/prometheus/src/main.rs
```diff
@@ -51,7 +51,12 @@ fn main() {
         .source(YellowstoneGrpcSource::new())
         .source(SolanaAccountsRpcSource::new())
         .account(Pipeline::new(RaydiumAmmV4AccParser, [Logger]))
-        .instruction(Pipeline::new(RaydiumAmmV4IxParser, [Logger]))
+        .instruction(Pipeline::new(
+            RaydiumAmmV4IxParser.filter()
+                .include_accounts(["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"])
+                .required_accounts([RaydiumAmmV4AccParser]),
```
kespinola:
```
.required_accounts([RaydiumAmmV4AccParser]),
```
I don't know about passing in a Parser to filter

kespinola:
```suggestion
        .instruction(FilteredPipeline::new(
            RaydiumAmmV4IxParser,
            [Preflilter::build().include_accounts(["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"]).unwrap()]
```

examples/prometheus/src/main.rs
```diff
@@ -51,7 +51,10 @@ fn main() {
         .source(YellowstoneGrpcSource::new())
         .source(SolanaAccountsRpcSource::new())
         .account(Pipeline::new(RaydiumAmmV4AccParser, [Logger]))
-        .instruction(Pipeline::new(RaydiumAmmV4IxParser, [Logger]))
+        .instruction(FilterPipeline::new(RaydiumAmmV4IxParser, [Logger])
+            .include_accounts(["GH8Ers4yzKR3UKDvgVu8cqJfGzU4cU62mTeg9bcJ7ug6", "4xxM4cdb6MEsCxM52xvYqkNbzvdeWWsPDZrBcTqVGUar"])
+            .required_accounts(["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"]),
+        )
```
kespinola:
```suggestion
        .instruction(
          FilterPipeline::new(
            RaydiumAmmV4IxParser, 
            [Logger], 
            Filter::build()
              .include_accounts([
                  "GH8Ers4yzKR3UKDvgVu8cqJfGzU4cU62mTeg9bcJ7ug6",
                  "4xxM4cdb6MEsCxM52xvYqkNbzvdeWWsPDZrBcTqVGUar"
                ])
              .required_accounts(["TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"])
          ),
        )
```

@fernandodeluret sorry I didnt mention this yesterday during our sync but looking more at the PR we are re-implementing include_accounts and required_accounts on the pipeline. In order to decouple concerns so more filter options can be added without having to adjust FIlterPipeline I think we should take a filter as the third-arguement to FilterPiplene then in its preflight it merges the filter on the parser with this extra filter. I believe Filter already supports merge.

kespinola (changes requested):

crates/runtime/src/buffer.rs
```diff
@@ -83,11 +83,18 @@ impl<M: Instrumenter, H: Send> topograph::AsyncHandler<Job, H> for Handler<M> {
...
+                        .run(span.clone(), &t, counters);
+
+                let instruction_fut = pipelines
+                    .instruction
                     .get_handlers(&filters)
-                    .run(span, &t, counters)
-                    .await;
+                    .run(span, &t, counters);
+
+                futures_util::future::join_all([transaction_fut, instruction_fut]).await;
```
kespinola, line 97: Can you take me through why we have 2 futures now?

kespinola (approved):
LG2M. Thank you for the last adjustment of FilterPipeline.

PR #105, opened 2025-07-13
Update Vixen parsers with last Codama updates

fernandodeluret:


crates/meteora-parser/src/custom_impl/oracle.rs
```diff
@@ -0,0 +1,34 @@
+use borsh::BorshDeserialize;
```
kespinola, line 1: What Ive been doing is adding extra to the lib.rs that way the generated mod is always codama generated but can add more to the lib.rs.

crates/meteora-parser/src/generated_parser/accounts_parser.rs
```diff
@@ -244,6 +244,11 @@ mod proto_parser {
                 idx: self.idx,
                 active_size: self.active_size,
                 length: self.length,
+                observations: self
+                    .observations
+                    .into_iter()
+                    .map(IntoProto::into_proto)
+                    .collect(),
```
kespinola: Is this field in the IDL?

fernandodeluret: no, this is not in the idl

PR #111, opened 2025-07-29
Add fumarole `Source`

fernandodeluret:
Adds Fumarole and bumps grpc version (to maintain compatibility between Filters implementations)

### DIFF
--- a/crates/yellowstone-fumarole-source/src/lib.rs
+++ b/crates/yellowstone-fumarole-source/src/lib.rs
@@ -1,0 +1,114 @@
+use std::collections::{BTreeMap, HashMap};
+
+use async_trait::async_trait;
+use bytesize::ByteSize;
+use tokio::{sync::mpsc::Sender, task::JoinSet};
+use yellowstone_fumarole_client::{DragonsmouthAdapterSession, FumaroleClient};
+use yellowstone_grpc_proto::{geyser::SubscribeUpdate, tonic::Status};
+use yellowstone_vixen::{config::YellowstoneConfig, sources::Source, Error as VixenError};
+use yellowstone_vixen_core::Filters;
+
+/// A `Source` implementation for the Yellowstone gRPC API.
+#[derive(Debug, Default)]
+pub struct YellowstoneFumaroleSource {
+    config: Option<FumaroleConfig>,
+    filters: Option<Filters>,
+}
+
+#[derive(Debug, Clone)]
+pub struct FumaroleConfig {
+    base: YellowstoneConfig,
+    /// Name of the persistent subscriber to use
+    subscriber_name: String,
+}
+
+impl From<FumaroleConfig> for yellowstone_fumarole_client::config::FumaroleConfig {
+    fn from(config: FumaroleConfig) -> Self {
+        yellowstone_fumarole_client::config::FumaroleConfig {
+            endpoint: config.base.endpoint,
+            x_token: config.base.x_token,
+            max_decoding_message_size_bytes: 512_000_000,
+            x_metadata: BTreeMap::new(),
+            response_compression: None,
+            request_compression: None,
+            initial_connection_window_size: ByteSize::mb(100),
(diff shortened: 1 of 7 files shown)


crates/yellowstone-fumarole-source/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        self.config = Some(FumaroleConfig {
+            base: config,
+            subscriber_name: "default_subscriber".to_string(),
+        });
+    }
+
+    fn get_filters(&self) -> &Option<Filters> { &self.filters }
+
+    fn get_config(&self) -> Option<YellowstoneConfig> {
+        self.config.as_ref().map(|config| config.base.clone())
```
kespinola: Why do we have these setter and getters?

Ideally like to pass all required arguments in at new but I can see how that but I can see why we need this. A side note but I'd like talk about the this. Not a blocker but curious.

crates/yellowstone-fumarole-source/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+            let dragonsmouth_session = fumarole_client
+                .dragonsmouth_subscribe(subscriber_name, filter.into())
+                .await
+                .expect("failing to subscribe to fumarole");
+
+            let DragonsmouthAdapterSession {
+                sink: _,
+                mut source,
+                fumarole_handle: _,
+            } = dragonsmouth_session;
```
lvboudre: `dragonsmouth_subscribe_with_config`
allows you to pass custom `FumaroleSuscribeConfig` which allow you to increase the number of tcp connection.

crates/yellowstone-fumarole-source/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn from(config: FumaroleConfig) -> Self {
+        let base_config = config.base.expect("FumaroleConfig.base is required");
+
+        yellowstone_fumarole_client::config::FumaroleConfig {
+            endpoint: base_config.endpoint,
+            x_token: base_config.x_token,
+            max_decoding_message_size_bytes: 512_000_000,
+            x_metadata: BTreeMap::new(),
+            response_compression: None,
+            request_compression: None,
```
lvboudre: You forgot to :`base_config.response_compression` and `base_config.request_compression`
This is important for firehose-type subscription

kespinola:
```
Instead of dragonsmouth_subscribe use dragonsmouth_subscribe_with_config and pass a custom FumaroleSubscribeConfig where num_data_plane_tcp_connections  is set to 4
```

@fernandodeluret give these comments a look. We need to adjust the config to get better performance.

crates/yellowstone-fumarole-source/src/lib.rs
fernandodeluret: Thanks for the advice! Will implement! :heart:

crates/yellowstone-fumarole-source/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+        for (filter_id, prefilter) in filters.parsers_filters {
+            let mut filter = Filters::new(HashMap::from([(filter_id, prefilter)]));
+            filter.global_filters = filters.global_filters;
+            let config = config.clone();
+            let subscriber_name = subscriber_name.clone();
+            let tx = tx.clone();
+            let fumarole_subscribe_config = FumaroleSubscribeConfig {
+                num_data_plane_tcp_connections: NonZero::new(filters_count).unwrap(),
+                ..Default::default()
+            };
```
kespinola: This should always be 4. That is based on benchmarking from the Triton team.

kespinola: How I see we have a max but lets always make it 4. Simple and hardcoded to start :)

Issue #101, opened 2025-07-03
Source connnection error should return run error


kespinola:
## Issues
On `YellowstoneGrpcSource` if can't connect to endpoint the runtime exits on `try_run_async` but the result is Ok so user is unable to handle the error case and close out appropriately.

## Example

In this example you see the error log from the failed connection. The snapshot future exits but its returns Ok(_) and does not trigger the Err arm.

```
2025-07-03T08:58:19.381543Z DEBUG rustls::webpki::anchors: add_parsable_certificates processed 150 valid and 0 invalid certs
2025-07-03T08:58:19.382006Z DEBUG hyper_util::client::legacy::connect::http: connecting to 10.43.64.244:28740

thread 'tokio-runtime-worker' panicked at /Users/kyleespinola/.cargo/git/checkouts/yellowstone-vixen-c8f39870d4021506/4c542b2/crates/runtime/src/lib.rs:380:21:
Source connection failed for: yellowstone-grpc
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-07-03T08:58:19.466279Z  WARN yellowstone_vixen::buffer: Server stopped sending updates
```

```rust
    let snapshot = Runtime::builder()
        .source(YellowstoneGrpcSource::new())
        .account(Pipeline::new(
            AccountParser::new(config.programs.into()),
            [SnapshotHandler::new(copy_in_sender)],
        ))
        .build(config.grpc.into())
        .try_run_async();

    tokio::select! {
        result = snapshot => match result {
            Ok(_) => {
                log::info!("Snapshot completed successfully.");
            },
            Err(e) => {
                log::error!("Snapshot encountered an error: {:?}", e);
                copy_in.abort()?;
                return Err(e.into());
            },
        },
```

PR #112, opened 2025-07-31
`feat` Return runtime error on stream error

fernandodeluret:
closes: #101

### DIFF
--- a/crates/runtime/src/buffer.rs
+++ b/crates/runtime/src/buffer.rs
@@ -8,7 +8,7 @@ use topograph::{
 use tracing::warn;
 use yellowstone_grpc_proto::{
     geyser::{subscribe_update::UpdateOneof, SubscribeUpdate, SubscribeUpdatePing},
-    tonic::Status,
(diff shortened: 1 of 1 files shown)


crates/runtime/src/buffer.rs
```diff
@@ -31,10 +31,24 @@ impl Buffer {
...
+            Err(e) => return Err(crate::Error::Io(std::io::Error::from(e))),
+        };
+
+        match result {
+            Ok(_stop_code) => Ok(()),
+            Err(e) => match e {
+                crate::Error::YellowstoneStatus(e) => {
+                    if e.code() == Code::Internal {
+                        tracing::error!("Yellowstone server credentials error");
+                    }
```
kespinola: Is code internal always credential errors? I'd just return the Error like you are and the caller can handle as they wish.

kespinola (changes requested):
LG2M just comment about removing the trace or making it more generic depending on code internal.

PR #115, opened 2025-08-07
Make vixen.toml config file dynamically contain sources config

fernandodeluret:


crates/yellowstone-fumarole-source/src/lib.rs
```diff
@@ -50,26 +60,63 @@ impl From<FumaroleConfig> for yellowstone_fumarole_client::config::FumaroleConfi
...
+                        &metrics_endpoint,
+                        metrics,
+                        None,
+                    ) {
+                        tracing::error!("Failed to push Fumarole metrics: {e:?}");
+                    }
+                })
+                .await;
+            }
+        })
```
kespinola: This is designed where if the binary using vixen when it has prometheus feature active on fumarole will export its metrics to the endpoint of the app. We don't need to expose our own. 

This is how I'd like our metrics to work too. We just support prometheus as a feature with no push gateway. The user has to setup their own server.

crates/yellowstone-grpc-source/src/lib.rs
```diff
@@ -28,10 +27,16 @@ impl YellowstoneGrpcSource {
...
+        raw_config: toml::Value,
+    ) -> Result<(), VixenError> {
         // We require that config and filters are set before connecting to the `Source`
         let filters = self.filters.clone().ok_or(VixenError::ConfigError)?;
-        let config = self.config.clone().ok_or(VixenError::ConfigError)?;
+
+        let config: YellowstoneConfig = serde::Deserialize::deserialize(raw_config)
+            .expect("Failed to deserialize YellowstoneConfig");
 
         let timeout = Duration::from_secs(config.timeout);
```
kespinola: Can `Source::new(config: toml::Value, filters:: Filters)` 

Then the Source trait can just be connect.

examples/streams-tracing/src/main.rs
```diff
@@ -85,7 +85,7 @@ async fn main() {
 
     let result = vixen::stream::Server::builder()
         .source(YellowstoneGrpcSource::new())
```
kespinola:
```suggestion
let result = vixen::stream::Server::builder()
.source::<YellowstoneGrpcSource>()
```

To get where new is invoked internally but call new once the runtime has gathered the filters and we read the connection config we call `T::new(config, filters)`

kespinola (changes requested):
I like where this is going. Good use of serde. We do loose the early typing but it would be a mess to make sources extensible and dynamic with static config.

We still cast it within the new. Ideally we would call new on all the sources and then connect. This allows the config to be validated without starting any connections to sources.

crates/yellowstone-grpc-source/src/lib.rs
fernandodeluret: The main problem is that we can't have a `new()` method in the Source trait (because it's not dyn compatible). I have been trying some workarounds for this, but very quickly it scalates the complexity and none looks clean at all. 

Regarding filters, we could receive them also in `connect()` like we are doing with config, but we would loose the ability to override filters for a particular source

PR #116, opened 2025-08-09
Remove setters and getters from SourceTrait

kespinola:
## Changes
- The SourceTrait now defines its config type as part of the trait definition. 
- Add DynSource which keeps consistent interface for starting a source. 
- Refactor builder to take .source<SourceTrait>::().


crates/runtime/src/lib.rs
```diff
@@ -373,23 +374,25 @@ impl<M: MetricsFactory> Runtime<M> {
...
 
-        for mut source in self.sources.drain(..) {
+        for source in self.sources.drain(..) {
             let tx = tx.clone();
-            source.filters(filters.clone());
-
+            let config = self
+                .sources_cfg
+                .get(&source.name())
+                .expect("No source config");
```
fernandodeluret, line 382: Very likely way too nitpicky comment but I maybe mention the source name in the error would make it easier to find out which source is failing. Here before we had this:
```
  "Source config not found for {source_name}. This should be set in your vixen \
   toml config file in the `[source.{source_name}]` section."
```

fernandodeluret (approved):

PR #118, opened 2025-08-14
Update Vixen metrics to receive prometheus register

fernandodeluret:


crates/runtime/src/handler.rs
```diff
@@ -318,9 +318,8 @@ where I::Item: AsRef<str> + Send + 'm
         futures_util::future::join_all(self.get_pipelines().map(move |(f, h)| {
             h.handle(value)
                 .map(move |r| {
-                    if let Some(r) = JobResult::from_pipeline(&r) {
-                        metrics.inc_processed(T::TYPE, r);
-                    }
+                    metrics::increment_processed_updates(&r, update_type);
```
kespinola, line 328: Is the feature flag for this in the increment_processed_updates? What happens when no registry is passed? My thought with the cfg feature is this code is dropped from the binary when the feature is not active.

fernandodeluret: I didn't set any cfg yet so that we could decide with the code how to implement it. So far it's only calling the fn and incrementing the atomic int but that's it (if not register is passed). But let me set all this under a feature flag

fernandodeluret: https://github.com/rpcpool/yellowstone-vixen/pull/118/commits/0a68bff3479306aea752c4aebab360ae39184ed3 @kespinola

PR #120, opened 2025-08-15
Snapshot Source, Single Source, Stream Crate, Slot Parser, 

kespinola:


Vixen.example.toml
```diff
@@ -1,7 +1,8 @@
 # This is an example TOML configuration file for Yellowstone Vixen.
 # Copy to Vixen.toml and them modify as needed.
 
-[yellowstone]
+# Only needed if you are using the Yellowstone gRPC as a source.
+[sources.yellowstone-grpc]
```
fernandodeluret: because of the way we are deserializing in config I think this is going to now be `[source]`

crates/core/src/lib.rs
```diff
@@ -154,7 +157,9 @@ pub struct Prefilter {
     /// Filters for transaction updates.
     pub transaction: Option<TransactionPrefilter>,
     /// Filters for block meta updates.
-    pub(crate) block_meta: Option<BlockMetaPrefilter>,
+    pub block_meta: Option<BlockMetaPrefilter>,
+    /// Filters for slot updates.
+    pub slot: Option<SlotPrefilter>,
```
kespinola, line 172: I didn't look into this but can we use a bool or does Vixen want an opt? Ideally we do slot default false and then set to true. I think you can. I did it somewhere else after I did this.

crates/core/src/lib.rs
```diff
@@ -445,6 +465,7 @@ pub enum PrefilterError {
 #[must_use = "Consider calling .build() on this builder"]
 pub struct PrefilterBuilder {
     error: Option<PrefilterError>,
+    slots: Option<()>,
```
kespinola:
```suggestion
    slots: bool,
```

crates/runtime/src/builder.rs
```diff
@@ -48,60 +48,61 @@ pub enum BuilderError {
...
+    /// The account pipelines.
+    pub account: Vec<BoxPipeline<'static, AccountUpdate>>,
+    /// The transaction pipelines.
+    pub transaction: Vec<BoxPipeline<'static, TransactionUpdate>>,
+    /// The instruction pipelines.
+    pub instruction: Vec<BoxPipeline<'static, InstructionUpdate>>,
+    /// The block meta pipelines.
+    pub block_meta: Vec<BoxPipeline<'static, BlockMetaUpdate>>,
+    /// The slot pipelines.
+    pub slot: Vec<BoxPipeline<'static, SlotUpdate>>,
```
kespinola, line 63: I didn't look into this but how do we handle multiple block_meta and slot pipelines? We copy the slot and block meta sending a unique copy to each handler? That is good if so or we keep this as a single pipeline for each instead of a list.

crates/solana-snapshot-source/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+                    if owners.contains(&account_owner) {
+                        let _ = sync_tx.send(Event::AccountUpdate(SubscribeUpdateAccount {
+                            account: Some(SubscribeUpdateAccountInfo {
+                                pubkey: account.pubkey().to_bytes().to_vec(),
+                                lamports: account.lamports,
+                                owner: account.owner.to_bytes().to_vec(),
+                                executable: account.executable,
+                                rent_epoch: account.rent_epoch,
+                                data: account.data.to_vec(),
+                                write_version: 0,
```
kespinola, line 198: I'll look at this but a not we can increment this as we go through the snapshot.

README.md
```diff
@@ -55,8 +55,8 @@ pub struct Opts {
 #[derive(Debug)]
 pub struct Logger;
 
-impl<V: std::fmt::Debug + Sync> vixen::Handler<V> for Logger {
-    async fn handle(&self, value: &V) -> vixen::HandlerResult<()> {
+impl<V: std::fmt::Debug + Sync> yellowstone_vixen::Handler<V> for Logger {
+    async fn handle(&self, value: &V) -> yellowstone_vixen::HandlerResult<()> {
```
kespinola, line 59: The main of the example below needs to be adjusted. Overall please do a review of our rustdoc and md docs to make sure they are up to date. We dropped multi source, changed the API, and no longer have from_slot etc on the builder.

crates/runtime/src/builder.rs
fernandodeluret: Let me get back to you on this one :pray:

PR #126, opened 2025-09-10
feat : add block subscriber to vixen and fix few bugs

Nagaprasadvr:
1. add block subscriber pipeline
2. add accounts_include_all filter to prefilter
3. fix few bugs with geyser-grpc subscription for slot and block_meta


crates/yellowstone-grpc-source/src/lib.rs
```diff
@@ -60,6 +60,8 @@ impl SourceTrait for YellowstoneGrpcSource {
 
             let mut client = GeyserGrpcClient::build_from_shared(config.endpoint.clone())?
                 .x_token(config.x_token.clone())?
+                .max_decoding_message_size(usize::MAX)
+                .accept_compressed(CompressionEncoding::Zstd)
```
kespinola: Since we have per source config now these are good to set through the config and can make the default the previous values we hit while testing.

kespinola (changes requested):

PR #128, opened 2025-09-24
`feat` Add metadata size to AccountsFile deserialization

fernandodeluret:
- Deserialize snapshot metadata file
- Use AccountFile lenght contained in metadata for deserialization

### DIFF
--- a/crates/solana-snapshot-source/src/lib.rs
+++ b/crates/solana-snapshot-source/src/lib.rs
@@ -1,8 +1,27 @@
-use std::{fs::File, path::PathBuf, sync::mpsc};
+#![allow(deprecated)]
+
+use std::{
+    collections::{HashMap, HashSet},
+    fs::File,
+    io::BufReader,
+    path::PathBuf,
+    sync::mpsc,
+};
 
 use async_trait::async_trait;
+use bincode::Options;
 use serde::{Deserialize, Serialize};
-use solana_accounts_db::accounts_file::AccountsFile;
+use solana_accounts_db::{
+    accounts_file::AccountsFile,
+    accounts_hash::{SerdeAccountsDeltaHash, SerdeAccountsHash},
+    ancestors::AncestorsForSerialization,
+    blockhash_queue::BlockhashQueue,
+};
+use solana_hash::Hash;
+use solana_program::clock::{Epoch, Slot, UnixTimestamp};
+use solana_pubkey::Pubkey;
+use solana_runtime::bank::BankHashStats;
+use solana_serde::default_on_eof;
 use tar::Archive;
 use tempfile::{tempdir, TempDir};
 use tokio::task::JoinSet;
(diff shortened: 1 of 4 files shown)


crates/solana-snapshot-source/src/lib.rs
```diff
@@ -42,21 +68,76 @@ impl SolanaSnapshot {
...
+
+                let (slot_str, id_str) = file_name
+                    .split_once('.')
+                    .unwrap_or_else(|| panic!("Invalid file name: {}", file_name));
+                let slot = slot_str
+                    .parse::<u64>()
+                    .unwrap_or_else(|_| panic!("Invalid slot: {}", slot_str));
+                let id = id_str
+                    .parse::<u64>()
+                    .unwrap_or_else(|_| panic!("Invalid id: {}", id_str));
```
kespinola, line 116: Can share details on this? Wonderful addition of write version. Please share any resources too.

crates/solana-snapshot-source/src/lib.rs
```diff
@@ -42,21 +68,76 @@ impl SolanaSnapshot {
...
+                    .get(&slot)
+                    .unwrap_or_else(|| panic!("accounts_metadata not found for slot: {}", slot));
+
+                let mut size = None;
+                for account in accounts_metadata {
+                    if account.id as u64 == id {
+                        size = Some(account.accounts_current_len);
+                        break;
+                    }
+                }
```
kespinola, line 128: Can you tell me a bit about this?

crates/solana-snapshot-source/src/lib.rs
```diff
@@ -113,7 +195,7 @@ impl SourceTrait for SolanaSnapshotSource {
                 },
             );
 
-        let solana_snapshot = SolanaSnapshot::unpack_compressed(config.path.clone())?;
+        let solana_snapshot = SolanaSnapshot::unpack_compressed(config.path.clone(), config.slot)?;
```
kespinola, line 198: If the slot is in the name of the file why pass it in? And also assuming folks don't change the filename not a big one but it is something that could break this. I don't think it should rely on the name of the tar but can rely on the naming and folder structure of the contents of the tar.

crates/solana-snapshot-source/src/lib.rs
fernandodeluret: There is not much resources about any of the changes here. I basically had to read through agave code, in theory we should only have 1 write version of the account per slot in snapshots (so also means 1 file per slot). But it's there just for cover that edge case. Also because it's a good way to double check that metadata deserialization went correctly (because we look on metadata not only by slot but also by write version)

crates/solana-snapshot-source/src/lib.rs
fernandodeluret: Well the basic idea of this(and all this change in general) is that we can teorically have files for which the accounts data does not match the file length (because of padding). So the safe way to decode accounts is to read the "accounts_current_len" from the metadata file, which is how agave is loading snapshots. In the tests I ran after the deserialization in practice they do match most of the time, but I left it there because it's the way that agave uses and also we cover the case there is difference between account data and file size

crates/solana-snapshot-source/src/lib.rs
fernandodeluret: This is true, actually for now this is draft, but I havethe intention of doing the slot optional, and if not passed use the metadata slot (the file name can also be checked). But this is a good double check (actually triple check) about the slot, because we can set as the slot in the config the data received as result from the endpoint we get the snapshot (and over which we do some calculations). So when dealing with several snapshots in the same process this is just an reduntant check

fernandodeluret: I still want to test further all this though (seems is working good so far, and could be a good resource for people working with snapshots, because there is not much docs out there about all the length check, at least that I have found)

PR #131, opened 2025-09-26
Move crates dependencies to workspace dependencies + unpinning

lvboudre:
The goal of moving deps at the workspace level it forces all crates the same lib version if needed.
In otherwords when you upgrade a lib, it forces software dev to migrated all crates too and not forgetting about any one of them, which is typically a desirable outcome.

As for Unpinning crates (using caret ^), this will allow other user of your crates to apply patches to common libraries they share with you. When we pin like "solana-pubkey = 2.2.1" we cannot benefits from say anza publishing bugfix/security fix of 2.2.3, etc.

Finally I added `protobuf-src` to various parser that needs to compile proto files as it offers better portability across development environment.


crates/boop-parser/Cargo.toml
```diff
@@ -1,33 +1,32 @@
-
-
 [package]
 name = "yellowstone-vixen-boop-parser"
-version = "0.5.0"
+version = "0.5.1"
```
kespinola, line 3: @lvboudre I use cargo release which handles incrementing the version based on patch, minor, major semantics. 

It will release without it so this fine but I've found cargo release helpful for workspace projects just FYI.

crates/boop-parser/Cargo.toml
```diff
@@ -1,33 +1,32 @@
...
 edition = "2021"
 description = "Vixen program parser for Boop program"
-license = "MIT"
-repository = "https://github.com/rpcpool/yellowstone-vixen"
+license.workspace = true
+repository.workspace = true
+authors.workspace = true
```
kespinola, line 8: Nice 🔥

crates/boop-parser/Cargo.toml
lvboudre: I was not aware of this thanks.
Next time I will know, won't do it this time since I have to undo in like 20 crates :p

lvboudre:
> I like the change about "protobuf_src", protoc was a source of some trouble in the past. Is this something that can increase too much compile times?

Yes it increase compilation time.

> Also regarding making the rust version fixed, I don't really have an opinion about it (it has pros and cons imo). But why is the main reason for it?

We use the rust-toolchain file in CI/CD, this is the common approach we used in other repo we have.
It can be improve in the future, but as a matter of consistency + fixing the ci/cd I went with fixing the rustc version.

PR #141, opened 2025-10-05
Fix: SingleInstructionPipeline early return prevents processing subsequent instructions

a00012025:
## Problem

In `SingleInstructionPipeline::handle()`, when a handler returns an error while processing instruction N, the method immediately returns, preventing subsequent instructions (N+1, N+2, ...) in the same transaction from being processed.

**Root Cause**: Early return in error handling path at `runtime/src/instruction.rs:113`

```rust
Err(e) => {
    let handled = e.handle::<InstructionUpdate>(&pipe.id());
    return Err(PipelineErrors::AlreadyHandled(handled));  // ← Stops loop
}
```

This causes data loss when handlers encounter recoverable errors (channel backpressure, validation errors, etc.).

## Solution

Match the behavior of `InstructionPipeline::handle()`: store errors but continue processing all instructions.

```rust
let mut err = None;
for insn in ixs.iter().flat_map(|i| i.visit_all()) {
    match res {
        Ok(()) => (),
        Err(PipelineErrors::AlreadyHandled(h)) => h.as_unit(),
        Err(e) => err = Some(e.handle::<InstructionUpdate>(&pipe.id())),  // Store, continue
    }
}
if let Some(h) = err {
    Err(PipelineErrors::AlreadyHandled(h))
} else {
    Ok(())
}
```

## Impact

- **Before**: Error on instruction N skips all remaining instructions in transaction
- **After**: All instructions processed; errors still logged and returned

This only affects pipelines created via `Runtime::builder().instruction()`. Manual `InstructionPipeline` creation was unaffected.

### DIFF
--- a/crates/runtime/src/instruction.rs
+++ b/crates/runtime/src/instruction.rs
@@ -95,6 +95,7 @@ impl SingleInstructionPipeline {
     /// # Errors
     /// Returns an error if the inner pipeline fails.
     pub async fn handle(&self, txn: &TransactionUpdate) -> Result<(), PipelineErrors> {
+        let mut err = None;
         let ixs = InstructionUpdate::parse_from_txn(txn).map_err(PipelineErrors::parse)?;
         let pipe = &self.0;
 
@@ -107,15 +108,15 @@ impl SingleInstructionPipeline {
             match res {
                 Ok(()) => (),
                 Err(PipelineErrors::AlreadyHandled(h)) => h.as_unit(),
-                Err(e) => {
-                    let handled = e.handle::<InstructionUpdate>(&pipe.id());
-
-                    return Err(PipelineErrors::AlreadyHandled(handled));
-                },
+                Err(e) => err = Some(e.handle::<InstructionUpdate>(&pipe.id())),
             }
         }
 
-        Ok(())
+        if let Some(h) = err {
+            Err(PipelineErrors::AlreadyHandled(h))
+        } else {
+            Ok(())
+        }
     }
 }
 


crates/runtime/src/instruction.rs
```diff
@@ -107,15 +108,15 @@ impl SingleInstructionPipeline {
             match res {
                 Ok(()) => (),
                 Err(PipelineErrors::AlreadyHandled(h)) => h.as_unit(),
-                Err(e) => {
-                    let handled = e.handle::<InstructionUpdate>(&pipe.id());
-
-                    return Err(PipelineErrors::AlreadyHandled(handled));
-                },
```
kespinola, line 114: If there is any instruction error then the whole transaction is aborted. This error handling is  designed to mimic the same transaction handling as the svm runtime. Is this not the cases?

kespinola (commented):
Thanks for thoughtful pull request. Let's discuss some.

crates/runtime/src/instruction.rs
a00012025: I think it happens when one handler returned an error when parsing an instruction, even that instruction succeeds. I notice this issue because it didn't parse the second jupiter route instruction in this tx: https://solscan.io/tx/3YEYmqJeRJHDNX4B7yY5mmjeGbT5scMw977LXQfC4CBTihX7qoSW3AQiZnFD8Us5jz2LXoTSu5HPF5XrfqzBcWUZ

a00012025: It's because I returned `yellowstone_vixen_core::ParseError::Filtered` in some of my parsers which will be treated as error here

kespinola: @a00012025 I haven't gotten back to this but I am hesitant. What if you handle the error in your parser and don't return an error unless its something you want to fail?

I do feel as an error in parsing any instruction in the transaction should early exit but lets keep talking about it.

PR #147, opened 2025-10-21
Add Raw Event To Handler Callback

CanardMandarin:
Hello,

A current issue with the handler callback is that it's not possible to access the raw event:

```rs
impl Handler<TokenProgramState> for TokenProgramAccountTransformer {
    async fn handle(&self, value: &TokenProgramState) -> HandlerResult<()> {
        /// Impossible to get the PUBKEY of the account update
        Ok(())
    }
}
```

Here is what you can do now:

```rs
impl Handler<TokenProgramState, AccountUpdate> for TokenProgramAccountTransformer {
    async fn handle(&self, value: &TokenProgramState, raw_event: &AccountUpdate) -> HandlerResult<()> {

        let account = match raw_event.account.as_ref() {
            Some(account) => account,
            None => return Ok(()),
        };
        let pubkey = Pubkey::try_from(&account.pubkey[..])?;
        let owner = Pubkey::try_from(&account.owner[..])?;
        let slot = raw_event.slot;
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,8 @@ pub struct Opts {
 #[derive(Debug)]
 pub struct Logger;
 
-impl<V: std::fmt::Debug + Sync> yellowstone_vixen::Handler<V> for Logger {
-    async fn handle(&self, value: &V) -> yellowstone_vixen::HandlerResult<()> {
+impl<V: std::fmt::Debug + Sync, R: Sync> vixen::Handler<V, R> for Logger {
+    async fn handle(&self, _value: &V, _raw: &R) -> vixen::HandlerResult<()> {
         tracing::info!(?value);
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


kespinola:
@CanardMandarin looking at the I'm wondering if we should parse the raw_event some so you don't have to deserialize the pubkeys and pull the account out of the option. Lets roll with this for now but potential future improvement.

kespinola (changes requested):
@CanardMandarin can you please update the readme handler preview so everything is up-to-date?

```rust
yellowstone_vixen::Runtime<YellowstoneGrpcSourc>::builder()
```

There is also a small typo in the source if you don't mind adding the "e" to `YellowstoneGrpcSource`

CanardMandarin:
> @CanardMandarin can you please update the readme handler preview so everything is up-to-date?
> 
> ```rust
> yellowstone_vixen::Runtime<YellowstoneGrpcSourc>::builder()
> ```
> 
> There is also a small typo in the source if you don't mind adding the "e" to `YellowstoneGrpcSource`

I've just updated the readme :)!

> @CanardMandarin looking at the I'm wondering if we should parse the raw_event some so you don't have to deserialize the pubkeys and pull the account out of the option. Lets roll with this for now but potential future improvement.

Indeed that would be more convenient, I'll see how we can do that!

PR #151, opened 2025-10-28
feat: Add Jetstreamer source for historical data replay (#149)

### LINKED ISSUE #149: 📡 Feature: Add Jetstreamer as a Data Source to Vixen
## 🧩 Context

[Jetstreamer](https://github.com/anza-xyz/jetstreamer) is a high-throughput toolkit for streaming historical Solana ledger data from the [Old Faithful](https://anza.xyz/old-faithful) archive. It can replay transactions and blocks from any epoch or slot range at >2.7M TPS and supports multi-threaded, network-based backfill.

While Jetstreamer does not currently include account updates or transaction logs, it is ideal for streaming blocks, transactions, rewards, and epochs into analytics or processing pipelines.

## 🎯 Objective

Integrate Jetstreamer as a new JetstreamSource for Vixen, allowing slot-range or epoch-based historical replay into the Vixen pipeline.

## ✅ Acceptance Criteria
- New Jetstream source module integrated with Vixen.
- Supports slot range or epoch input for backfill.
- Emits transactions, blocks, rewards, and epochs into the Vixen pipeline.
- Handles multithreaded replay and out-of-order slot delivery.
- Includes an example or test for replaying a known epoch (e.g., epoch 800).

## 📎 Resources
- GitHub: https://github.com/anza-xyz/jetstreamer
- Docs: https://docs.rs/jetstreamer
- Archive: https://anza.xyz/old-faithful

senzenn:
## Summary
Adds Jetstreamer as a new data source for replaying historical Solana ledger data from Old Faithful archive.

Closes #149

## Changes
- ✅ New `jetstreamer-source` crate with `SourceTrait` implementation
- ✅ Epoch and slot-range based queries
- ✅ Multi-threaded processing with out-of-order slot handling
- ✅ Example: `jetstream-replay` for epoch 800
- ✅ Updated `Vixen.example.toml` with Jetstream config
- ✅ Unit tests and documentation

## Usage
```toml
[source]
archive_url = "https://api.old-faithful.net"
epoch = 800
threads = 4
```

```bash
cargo run --example jetstream-replay
```

## Requirements
⚠️ Requires Rust 1.88+ (due to `jetstreamer-utils` dependency)

## Acceptance Criteria
- [x] Jetstream source integrated with Vixen
- [x] Supports slot range and epoch input
- [x] Emits blocks/transactions into pipeline
- [x] Handles multi-threaded replay and out-of-order slots
- [x] Includes epoch 800 replay example

## Testing
```bash
cargo test --package jetstreamer-source
cargo run --example jetstream-replay
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,9 +65,9 @@ dependencies = [
 
 [[package]]
 name = "agave-feature-set"
-version = "2.3.10"
+version = "2.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "98b7122392ed81e9b4569c3c960a557afb9f735719e0bcc9b2c19d0345568f5f"
+checksum = "d52a2c365c0245cbb8959de725fc2b44c754b673fdf34c9a7f9d4a25c35a7bf1"
 dependencies = [
  "ahash 0.8.12",
  "solana-epoch-schedule 2.2.1",
@@ -108,9 +108,9 @@ dependencies = [
 
 [[package]]
 name = "agave-io-uring"
-version = "2.3.10"
+version = "2.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e9a7a05ffb5fb07af5b5ce53c8206f881b118ac3d7aee52d6a3e0329b6d861a"
+checksum = "a10b918a355bc78764aceb688dbbb6af72425f62be9dbfb7beb00b6d3803a0bd"
 dependencies = [
  "io-uring",
  "libc",
@@ -162,9 +162,9 @@ dependencies = [
 
 [[package]]
 name = "agave-reserved-account-keys"
-version = "2.3.10"
+version = "2.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "44941ddf7f11f385b8a10a95bdbe8026909938a4932f23ccd2c78fc2a480a8d7"
+checksum = "8289c8a8a2ef5aa10ce49a070f360f4e035ee3410b8d8f3580fb39d8cf042581"
 dependencies = [
  "agave-feature-set 2.3.10",
  "solana-pubkey 2.4.0",
@@ -1195,6 +1195,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo_toml"
+version = "0.20.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "88da5a13c620b4ca0078845707ea9c3faf11edbc3ffd8497d11d686211cd1ac0"
+dependencies = [
+ "serde",
+ "toml 0.8.23",
+]
+
+[[package]]
 name = "cc"
 version = "1.2.39"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1369,6 +1379,59 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e7caa3f9de89ddbe2c607f4101924c5abec803763ae9534e4f4d7d8f84aa81f0"
 dependencies = [
  "cc",
+]
+
+[[package]]
+name = "codama-errors"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69a7b7b16f9b57be0ca6ea44b40f5e6b4aeedcd30bfd33d2998721dd9c2d807f"
+dependencies = [
+ "cargo_toml",
+ "proc-macro2",
+ "serde_json",
+ "syn 2.0.106",
+ "thiserror 2.0.16",
+]
+
+[[package]]
+name = "codama-nodes"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf3194539453ecd51fb72c8440e3aa4a84352a41d61dea8f37d1d5f2428d0419"
+dependencies = [
+ "codama-errors",
+ "codama-nodes-derive",
+ "derive_more",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "codama-nodes-derive"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c138c3a7ed16bdcea33cb7b708a650b575464d86077a558c3b6ce091d4ec7455"
+dependencies = [
+ "codama-errors",
+ "codama-syn-helpers",
+ "derive_more",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.106",
+]
+
+[[package]]
+name = "codama-syn-helpers"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a1ce2a27f09caf6476c857c809718fac9213ce4124ad0002f1f12f2964681ccc"
+dependencies = [
+ "codama-errors",
+ "derive_more",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.106",
 ]
 
 [[package]]
@@ -1868,34 +1931,23 @@ dependencies = [
 
 [[package]]
 name = "derive_more"
-version = "0.99.20"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6edb4b64a43d977b8e99788fe3a04d483834fba1215a7e02caa415b626497f7f"
+checksum = "4a9b99b9cbbe49445b21764dc0625032a89b145a2642e67603e1c936f5458d05"
 dependencies = [
- "convert_case",
+ "derive_more-impl",
+]
+
+[[package]]
+name = "derive_more-impl"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb7330aeadfbe296029522e6c40f315320aba36fc43a5b3632f3795348f3bd22"
+dependencies = [
  "proc-macro2",
  "quote",
- "rustc_version",
  "syn 2.0.106",
 ]
-
-[[package]]
-name = "dialoguer"
-version = "0.10.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59c6f2989294b9a498d3ad5491a79c6deb604617378e1cdc4bfc1c1361fe2f87"
-dependencies = [
- "console 0.15.11",
- "shell-words",
- "tempfile",
- "zeroize",
-]
-
-[[package]]
-name = "difflib"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6184e33543162437515c2e2b48714794e37845ec9851711914eec9d308f6ebe8"
 
 [[package]]
 name = "digest"
@@ -6414,9 +6466,9 @@ dependencies = [
 
 [[package]]
 name = "solana-account-decoder"
-version = "2.3.10"
+version = "2.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "902c42492c67e2cb78e839b1af063ed672b3d1ff22311ce07aa073804ca40e53"
+checksum = "ba71c97fa4d85ce4a1e0e79044ad0406c419382be598c800202903a7688ce71a"
 dependencies = [
  "Inflector",
  "base64 0.22.1",
@@ -6500,9 +6552,9 @@ dependencies = [
 
 [[package]]
 name = "solana-account-decoder-client-types"
-version = "2.3.10"
+version = "2.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "39260da5bed46e52afe17c8dc2fd41b7d23a946c18605154033631a334e882ec"
+checksum = "5519e8343325b707f17fbed54fcefb325131b692506d0af9e08a539d15e4f8cf"
 dependencies = [
  "base64 0.22.1",
  "bs58",
@@ -6556,9 +6608,9 @@ dependencies = [
 
 [[package]]
 name = "solana-accounts-db"
-version = "2.3.10"
+version = "2.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "97bb9be7b256b2295c443e054229942a34b40a2c032631d6e1c5a063ee1ed9e5"
+checksum = "dbbe35141711500d113dfc7aa79eb250c4458f04e759a67ba4bffc3e6cddc402"
 dependencies = [
  "agave-io-uring 2.3.10",
  "ahash 0.8.12",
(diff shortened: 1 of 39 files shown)


examples/jetstream-replay/src/main.rs
```diff
@@ -0,0 +1,146 @@
...
+    );
+
+    // Create empty filters
+    let filters = yellowstone_vixen_core::Filters::new(std::collections::HashMap::new());
+
+    // Create Jetstream source
+    let source = JetstreamSource::new(config, filters);
+
+    // Create channel for updates
+    let (tx, mut rx) = mpsc::channel(1000);
```
kespinola: Similar to the other examples can you wire up jet stream through vixen and then showcase on token program instructions?

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+        reorder_buffer: Arc<RwLock<ReorderBuffer>>,
+    ) -> Self {
+        Self { tx, reorder_buffer }
+    }
+
+    async fn process_block(&self, block: BlockData) -> Result<(), Error> {
+        info!(slot = block.slot(), "Processing block");
+
+        #[cfg(feature = "prometheus")]
+        crate::metrics::JETSTREAM_BLOCKS_RECEIVED.inc();
```
kespinola: Can you wrap in a method `jetstream_block_received()`

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+        });
+
+        Ok(())
+    }
+}
+
+impl JetstreamSource {
+    async fn stream_loop(
+        config: JetstreamSourceConfig,
+        _filters: Filters,
```
kespinola: We should be using the filter to only send updates on the transactions by owner that the user has configured. If they use use the block parser then we send them everything.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+                            parent_slot,
+                            parent_blockhash: parent_blockhash.to_string(),
+                        }
+                    )),
+                    created_at: Some(yellowstone_grpc_proto::prost_types::Timestamp::from(std::time::SystemTime::now())),
+                };
+
+                {
+                    let mut buffer = self.reorder_buffer.write().await;
+                    let ready_updates = buffer.add_slot(slot, vec![update]);
```
kespinola: Can you talk through the re-order buffer?

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+
+struct ReorderBuffer {
+    buffer: BTreeMap<u64, Vec<SubscribeUpdate>>,
+    next_expected_slot: u64,
+    max_size: usize,
+    timeout: Duration,
+    slot_timestamps: BTreeMap<u64, Instant>,
+}
+
+impl ReorderBuffer {
```
kespinola: You are using this to establish slot order from the threaded execution of jetstream? I don't think that is necessary at this level. If you do want to add a mechanism for ordering it should be done in the handler. You could offer an extension users can use to create an order buffer but in index cases its normal for them to write slot to a data store which can support out of order inserts.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+        let on_tx = Some(move |_thread_id: usize, tx: TransactionData| {
+            let handler = handler_clone.clone();
+            async move {
+                handler
+                    .process_transaction(tx)
+                    .await
+                    .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send>)
+            }
+            .boxed()
+        });
```
kespinola:
```suggestion
        let handler_on_tx = handler.clone();
        let on_tx = Some(move |_thread_id: usize, tx: TransactionData| {
            let handler_callback = handler_on_tx.clone();
            async move {
                handler
                    .process_transaction(tx)
                    .await
                    .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send>)
            }
            .boxed()
        });
```

A bit more expressive names for the handler clones.

examples/jetstream-replay/src/main.rs
senzenn: Updated jetstream-replay with Vixen integration and token parsing! #151

crates/jetstreamer-source/src/lib.rs
senzenn: Yes ! The reorder buffer ensures slots are delivered in correct order despite out-of-sequence arrival

crates/jetstreamer-source/src/lib.rs
senzenn:
```rust
// crates/jetstreamer-source/src/lib.rs
fn jetstream_block_received() {
    #[cfg(feature = "prometheus")]
    crate::metrics::JETSTREAM_BLOCKS_RECEIVED.inc();
}

fn jetstream_transaction_received() {
    #[cfg(feature = "prometheus")]
    crate::metrics::JETSTREAM_TRANSACTIONS_RECEIVED.inc();
}
```

crates/jetstreamer-source/src/lib.rs
senzenn:
```rust
// Check configured filters
for (filter_id, prefilter) in &self.filters.parsers_filters {
    if let Some(_tx_filter) = &prefilter.transaction {
        // NOTE: Full transaction data available via tx.transaction
        // Parse VersionedTransaction.message.account_keys for filtering

        if self.config.permissive_transaction_filtering {
            info!(filter_id = %filter_id, signature = ?tx.signature, "Processing transaction (permissive mode)");
            return true;
        } else {
            // Strict mode: Account key extraction not implemented yet
            info!(filter_id = %filter_id, signature = ?tx.signature, "Strict filtering requires account key extraction - not yet implemented");
            continue;
        }
    }

    // Process if filter has account/slot requirements
    if prefilter.account.is_some() || prefilter.slot.is_some() {
        debug!(filter_id = %filter_id, "Filter has account/slot requirements - processing transaction");
        return true;
    }
}
```

senzenn: We can improve this.

crates/jetstreamer-source/src/lib.rs
senzenn: Agree. I’ve removed the slot ordering logic from the source and left ordering to downstream handlers/storage. Jetstreamer can emit out-of-order across threads; consumers that require order can handle it in their handler or datastore.

crates/jetstreamer-source/src/lib.rs
senzenn: applied the naming improvement. 
```rust
         end_slot, "Starting Old Faithful streaming from {}", config.archive_url
        );

        let handler = Arc::new(VixenStreamHandler::new(
            tx.clone(),
            filters.clone(),
            config.clone(),
        ));

        std::env::set_var("JETSTREAMER_NETWORK", config.network.clone());
        std::env::set_var(
            "JETSTREAMER_COMPACT_INDEX_BASE_URL",
            config.compact_index_base_url.clone(),
        );
        std::env::set_var(
            "JETSTREAMER_NETWORK_CAPACITY_MB",
            config.network_capacity_mb.to_string(),
        );

```

crates/jetstreamer-source/src/metrics.rs
```diff
@@ -0,0 +1,37 @@
...
+        "Total transactions received from Jetstream",
+    ))
+    .unwrap()
+});
+
+/// Register all Jetstream metrics with the provided Prometheus registry
+pub fn register_metrics(registry: &Registry) {
+    let _ = registry.register(Box::new(JETSTREAM_BLOCKS_RECEIVED.clone()));
+    let _ = registry.register(Box::new(JETSTREAM_TRANSACTIONS_RECEIVED.clone()));
+}
```
kespinola: https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/runtime/src/metrics.rs#L250-L279

There is already metric for blocks and transactions received please remove from the source.

examples/jetstream-replay/config.toml
```diff
@@ -0,0 +1,11 @@
+[source]
+archive_url = "https://api.old-faithful.net"
+epoch = 800
+threads = 4
+reorder_buffer_size = 1000
+slot_timeout_secs = 30
```
kespinola: Can you please adjust serde attributes for dash case?

```suggestion
archive-url = "https://api.old-faithful.net"
epoch = 800
threads = 4
reorder-buffer-size = 1000
slot-timeout-secs = 30
```

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,569 @@
...
+
+    fn jetstream_block_received() {
+        #[cfg(feature = "prometheus")]
+        crate::metrics::JETSTREAM_BLOCKS_RECEIVED.inc();
+    }
+
+    fn jetstream_transaction_received() {
+        #[cfg(feature = "prometheus")]
+        crate::metrics::JETSTREAM_TRANSACTIONS_RECEIVED.inc();
+    }
```
kespinola: Metrics handled by the runtime you can drop them in the source.

https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/runtime/src/metrics.rs#L250-L279

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,569 @@
...
+    #[arg(long, env, default_value = "https://files.old-faithful.net")]
+    pub compact_index_base_url: String,
+
+    /// Network capacity in MB
+    #[arg(long, env, default_value = "1000")]
+    pub network_capacity_mb: usize,
+
+    /// Control transaction filtering: true = permissive (all), false = strict (limited).
+    #[arg(long, env, default_value = "true")]
+    pub permissive_transaction_filtering: bool,
```
kespinola: Should not have this option its confusing API. You can make a parser that matches all instead.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,569 @@
...
+                    >,
+                >,
+            >,
+            None,
+        )
+        .await;
+
+        if let Err((error, slot)) = result {
+            let error_msg = format!("{:?}", error);
+            if error_msg.contains("incomplete frame") {
```
kespinola, line 471: I little fragile since we can't match an error but ok.

kespinola (changes requested):

crates/jetstreamer-source/src/lib.rs
senzenn: fixed this in lts pr

crates/jetstreamer-source/src/lib.rs
senzenn: removed certain confusing api's in Lts Pr

Vixen.example.toml
```diff
@@ -21,6 +21,22 @@ timeout = 60
...
+slot-timeout-secs = 30
+network = "mainnet"
+compact-index-base-url = "https://files.old-faithful.net"
+network-capacity-mb = 1000
+permissive-transaction-filtering = true
+
+[filters]
+# Add filters for specific programs/accounts if needed
+# programs = ["11111111111111111111111111111112"]
+
```
kespinola, line 38: The filters should come from the parser's attached filters which mimic grpc subscribe filter notation for selecting account owner or address in accounts list for transactions we shouldn't need to specify filters on the source.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+                slot,
+                blockhash,
+                rewards,
+                block_time,
+                block_height,
+                executed_transaction_count,
+                entry_count,
+            } => {
+                let update = SubscribeUpdate {
+            filters: vec![],
```
kespinola: @senzenn Vixen uses this filter list to match a subscribe update with a parser so you need to include the filter key for Vixen to match on when it goes through the runtime.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+        );
+
+        // Check if transaction should be filtered
+        if !self.should_process_transaction(&tx) {
+            debug!(signature = ?tx.signature, slot = tx.slot, "Transaction filtered out");
+            return Ok(());
+        }
+        // TODO: Populate transaction field with protobuf-encoded transaction data
+        let update = SubscribeUpdate {
+            filters: vec![],
```
kespinola: Same here that want to use the filters to let Vixen know what parser and handler to take the event through.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+        if let Err(e) = self.tx.send(Ok(update)).await {
+            let error_msg = format!("Failed to send transaction update: {}", e);
+            error!("{}", error_msg);
+            return Err(Error::ChannelSend(error_msg));
+        }
+
+        Ok(())
+    }
+
+    fn should_process_transaction(&self, tx: &TransactionData) -> bool {
```
kespinola: Instead of returning boolean this should return Vec<String> to set on the filters key of the subscribe update. If its empty then you can toss out the event and don't send it to the channel.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+    #[arg(long, env, default_value = "1000")]
+    pub reorder_buffer_size: usize,
+
+    /// Slot timeout in seconds
+    #[arg(long, env, default_value = "30")]
+    pub slot_timeout_secs: u64,
+
+    /// Control transaction filtering: true = permissive (all), false = strict (limited).
+    #[arg(long, env, default_value = "true")]
+    pub permissive_transaction_filtering: bool,
```
kespinola:
```suggestion
```

I'd like to drop as filters are governed by the registered parsers so as long as there is 1 parser registered then it will match what is desired. I imagine this was helpful for testing through so understand why it was initially included.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+        let on_block = Some(move |_thread_id: usize, block: BlockData| {
+            let handler = handler_clone.clone();
+            async move {
+                handler
+                    .process_block(block)
+                    .await
+                    .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send>)
+            }
+            .boxed()
+        });
```
kespinola:
```suggestion
        let handler_on_block = handler.clone();
        let on_block = Some(move |_thread_id: usize, block: BlockData| {
            let handler_callback = handler_on_block.clone();
            async move {
                handler
                    .process_block(block)
                    .await
                    .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send>)
            }
            .boxed()
        });
```

crates/jetstreamer-source/src/metrics.rs
```diff
@@ -0,0 +1,8 @@
+//! Prometheus metrics for Jetstream source
+
+use prometheus::Registry;
+
+/// Register all Jetstream metrics with the provided Prometheus registry
+pub fn register_metrics(_registry: &Registry) {
+    // No metrics to register currently
+}
```
kespinola: Can delete this file.

crates/jetstreamer-source/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+yellowstone-grpc-proto = { workspace = true }
+yellowstone-grpc-client = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+clap = { workspace = true, features = ["derive", "cargo", "wrap_help"] }
+thiserror = { workspace = true }
+tokio-util = { workspace = true }
+prost-types = { workspace = true }
+# Jetstreamer firehose for direct streaming without ClickHouse
+jetstreamer-firehose = "0.1.6"
+jetstreamer-utils = "0.1.6"
```
kespinola: Can you include in the root Cargo.toml and then reference through the workspace so all versions are set in one place. We can also allow `0.1` to allow matching of any patches that come out.

examples/jetstream-replay/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+async-trait = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
+tracing = { workspace = true }
+tracing-subscriber = "0.3"
+anyhow = { workspace = true }
+yellowstone-vixen = { workspace = true, features = ["prometheus"] }
+yellowstone-vixen-core = { workspace = true }
+yellowstone-grpc-proto = { workspace = true }
+yellowstone-vixen-jetstream-source = { path = "../../crates/jetstreamer-source", features = ["prometheus"] }
+yellowstone-vixen-parser = { path = "../../crates/parser", features = ["token-program"] }
```
kespinola: Can these be referenced by workspace once set there?

examples/jetstream-replay/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+prometheus = { workspace = true }
+warp = "0.3"
```
kespinola: Include in Cargo.toml of root and reference through workspace.

PR #156, opened 2025-11-14
Fix Filter Preflight On Solana Snapshot Source

CanardMandarin:
Fix the Solana snapshot source filtering logic

### DIFF
--- a/crates/solana-snapshot-source/src/lib.rs
+++ b/crates/solana-snapshot-source/src/lib.rs
@@ -1,4 +1,9 @@
-use std::{collections::HashMap, fs::File, path::PathBuf, sync::mpsc};
+use std::{
+    collections::HashMap,
+    fs::File,
+    path::PathBuf,
+    sync::{mpsc, Arc},
+};
 
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
@@ -14,7 +19,7 @@ use yellowstone_grpc_proto::{
     tonic::Status,
 };
 use yellowstone_vixen::{sources::SourceTrait, Error as VixenError};
-use yellowstone_vixen_core::Filters;
+use yellowstone_vixen_core::{Filters, Pubkey};
 use zstd::Decoder;
 
 pub struct AccountFile(PathBuf, usize);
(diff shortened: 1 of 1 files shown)


crates/solana-snapshot-source/src/lib.rs
```diff
@@ -213,6 +220,6 @@ impl SourceTrait for SolanaSnapshotSource {
 }
 
 enum Event {
-    AccountUpdate(SubscribeUpdateAccount),
+    AccountUpdate(SubscribeUpdateAccount, Vec<String>),
     SnapshotFinished,
```
kespinola, line 234:
```suggestion
    AccountUpdate {
      account_update: SubscribeUpdateAccount,
      filters: Vec<String>
    },
SnapshotFinished,
```

crates/solana-snapshot-source/src/lib.rs
```diff
@@ -179,21 +177,30 @@ impl SourceTrait for SolanaSnapshotSource {
...
+                                    data: account.data.to_vec(),
+                                    write_version: 0,
+                                    txn_signature: None,
+                                }),
+                                slot,
+                                is_startup: true,
+                            },
+                            filter_keys,
+                        ));
                     }
```
kespinola:
```suggestion
                    if let Some(filter)  = filter_owner_key_lookup.lookup_by_owner(account_owner) {

                        let _ = sync_tx.send(Event::AccountUpdate {
                            account_update: SubscribeUpdateAccount {
                                account: Some(SubscribeUpdateAccountInfo {
                                    pubkey: account.pubkey().to_bytes().to_vec(),
                                    lamports: account.lamports,
                                    owner: account.owner.to_bytes().to_vec(),
                                    executable: account.executable,
                                    rent_epoch: account.rent_epoch,
                                    data: account.data.to_vec(),
                                    write_version: 0,
                                    txn_signature: None,
                                }),
                                slot,
                                is_startup: true,
                            },
                            filter,
                        });
}
```

crates/solana-snapshot-source/src/lib.rs
```diff
@@ -95,23 +95,23 @@ impl SourceTrait for SolanaSnapshotSource {
...
+                    owner_to_filters
+                        .entry(*owner)
+                        .or_default()
+                        .push(key.clone());
+                    if !all_owners.contains(owner) {
+                        all_owners.push(*owner);
+                    }
+                }
+            }
+        }
```
kespinola:
```suggestion
  let filter_owner_key_lookup = FitlerOwnerKeyLookup::new(filters.parsers_filters);
```

In terms of code structure having `FilterOwnerKeyLookup` to encapsulate this logic would be nice.

```rust
pub struct FilterOwnerLookup(Arc<HashMap<yellowstone_vixen_core::Pubkey, String>));

impl FilterOwnerKeyLookup {
  fn new(parsers_filter: ) -> Self {
      let mut lookup = HashMap::new();
        // convert the parser filter into lookup hash
       Self(Arc::new(lookup))
  }

  fn key_by_owner(pubkey: Pubkey) -> Option<Sting> {
     // Check the hash if it doesn't have then return none 
    // If has then you clone a copy and return Some(key)
  }
}

```

kespinola (approved):

PR #157, opened 2025-11-16
bump pumpfun / pumpswaps to v2 (and add pump fees)

hack-wrench:
1. Actualized idl.json with new pumpfun features (https://github.com/pump-fun/pump-public-docs)
2. Deleted from idl.json `OptionBool` spec, because it breaking codama
3. Added pump-fees parser
4. Added .cjs scripts for generating code in futures

### DIFF
--- a/crates/pump-fees-parser/Cargo.toml
+++ b/crates/pump-fees-parser/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "yellowstone-vixen-pump-fees-parser"
-version = "0.5.2"
+version.workspace = true
 edition.workspace = true
 description = "Vixen program parser for Pump Fees program"
 license.workspace = true


hack-wrench:
Ok, ill try

hack-wrench:
DESCRIPTOR_SET is not exposed for generated code

kespinola:
> DESCRIPTOR_SET is not exposed for generated code

@hack-wrench we are likely to deprecate gRPC support. Is this something you are using?

> And personally, i find it much more convenient to see the code during debugging than to refer to the proc macro every time

Noted. The reason for the proc-macro was to make it easier to keep program parsers up to date when program changes. Now just need to replace the idl.json and remove the extra step of generating a parser crate but I get where you are coming from.

Do you think more documentation on expected output of the proc-macro would help with integration?

Closing issue since we are moving to proc-macro but happy to keep talking about improving the DX.

PR #160, opened 2025-11-22
feat: add multiple program subscriptions

sonicfromnewyoke:
### Problem

this is not so simple to subscribe to multiple programs

### Summary of Changes

- modify `SubscribeRequest` in `stream.proto` to accept a list of program addresses
- modify join logic itself
- add error context to the `Error` in the `runtime` to make errors more descriptive
- lower log level from `warn` to `trace` for **"No pipeline matched filter on incoming update"** message, to make execution less spamming
- add `.DS_Store` to `.gitignore`
- apply `cargo fmt` to rust files which were involved in completion of that change. Diff is huge because project isn't formatted at all :(

cc @kespinola

### DIFF
--- a/crates/stream/src/grpc.rs
+++ b/crates/stream/src/grpc.rs
@@ -13,7 +13,7 @@ use yellowstone_vixen_proto::{
     stream::{
         self,
         program_streams_server::{ProgramStreams, ProgramStreamsServer},
-        SubscribeRequest, SubscribeManyRequest, SubscribeUpdate,
+        SubscribeManyRequest, SubscribeRequest, SubscribeUpdate,
     },
     tonic::{self, transport, Request, Response, Status},
     tonic_reflection,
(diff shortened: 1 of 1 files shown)


kespinola:
> apply cargo fmt to rust files which were involved in completion of that change. Diff is huge because project isn't formatted at all :(

It is formatted with `cargo +nightly fmt`

You'll CI check right now :)

kespinola:
@sonicfromnewyoke I'm checking with RPC providers that are offering a hosted version of the Vixen stream since this presents a breaking change.

Is opening multiple subscribe requests difficult to manage? It creates a 1:1 relationship which makes handling the events straight forward IMO but I do see the advantage of subscribing to multiple programs too. I just don't think this is good DX until we can get the event output typed.

sonicfromnewyoke:
> @sonicfromnewyoke I'm checking with RPC providers that are offering a hosted version of the Vixen stream since this presents a breaking change.
> 
> Is opening multiple subscribe requests difficult to manage? It creates a 1:1 relationship which makes handling the events straight forward IMO but I do see the advantage of subscribing to multiple programs too. I just don't think this is good DX until we can get the event output typed.

merging is ok but tedious and feels like you are doing something wrong

Alternatevely we can**extend** the API and having `message SubscribeRequest` and `message SubscribeManyRequest` which should make the migration smooth enough. What do you think about it?

PR #169, opened 2025-12-04
fix panics on account[idx]

grooviegermanikus:
> thread 'tokio-runtime-worker' panicked at pump_amm.rs:7:1:
index out of bounds: the len is 11 but the index is 11

### DIFF
--- a/crates/proc-macro/src/render.rs
+++ b/crates/proc-macro/src/render.rs
@@ -355,8 +355,9 @@ fn render_instruction_parser(
                 .enumerate()
                 .map(|(idx, account)| {
                     let field = format_ident!("{}", to_snake_case(&account.name));
+                    let error_msg = format!("Account does not exist at index {idx}");
 
-                    quote! { #field: *accounts.get(#idx).ok_or(ParseError::from("idx error"))? }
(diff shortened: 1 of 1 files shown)


crates/proc-macro/src/render.rs
```diff
@@ -356,7 +356,7 @@ fn render_instruction_parser(
                 .map(|(idx, account)| {
                     let field = format_ident!("{}", to_snake_case(&account.name));
 
-                    quote! { #field: accounts[#idx] }
+                    quote! { #field: *accounts.get(#idx).ok_or(ParseError::from("idx error"))? }
```
kespinola:
```suggestion
.map(|(idx, account)| {
let field = format_ident!("{}", to_snake_case(&account.name));
let error_msg = format!("Account does not exist at index {?idx}");

                    quote! { #field: *accounts.get(#idx).ok_or(ParseError::from(#error_msg
                    ))? }
```

kespinola (changes requested):
Recommend more descriptive error message but epic!

crates/proc-macro/src/render.rs
grooviegermanikus: could not apply from GH web-ui . pushed a new commit

PR #173, opened 2025-12-09
integrate grpc autoconnect

grooviegermanikus:
add this config option (default=`simple`):
```toml
[source]
# simple/autoreconnect
strategy = "autoreconnect"
```

### DIFF
--- a/crates/yellowstone-grpc-source/Cargo.toml
+++ b/crates/yellowstone-grpc-source/Cargo.toml
@@ -12,7 +12,8 @@ readme = "./README.md"
 async-trait = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread", "signal"] }
 tokio-util = { workspace = true }
-tonic-health = "0.14.2"
+# align version with yellowstone-grpc-client -> cargo tree -i tonic-health
+tonic-health = "0.14"
 tracing = { workspace = true }
 futures = { workspace = true }
 futures-util = { workspace = true, features = ["sink"] }


crates/yellowstone-grpc-source/src/simple_connector.rs
```diff
@@ -0,0 +1,80 @@
...
+use yellowstone_vixen::{sources::SourceTrait, CommitmentLevel, Error as VixenError};
+use yellowstone_vixen_core::Filters;
+use crate::{YellowstoneGrpcConfig};
+
+/// A `Source` implementation for the Yellowstone gRPC API.
+#[derive(Debug)]
+pub struct YellowstoneGrpcSource {
+    filters: Filters,
+    config: YellowstoneGrpcConfig,
+}
```
kespinola: I'm up to have auto-connect be configured through config instead of separate source?

crates/yellowstone-grpc-source/src/simple/simple_connector.rs
```diff
@@ -0,0 +1,80 @@
...
+    async fn connect(&self, tx: Sender<Result<SubscribeUpdate, Status>>) -> Result<(), VixenError> {
+        let filters = self.filters.clone();
+        let config = self.config.clone();
+
+        let timeout = Duration::from_secs(config.timeout);
+
+        let mut tasks_set = JoinSet::new();
+
+        for (filter_id, prefilter) in filters.parsers_filters {
+            let filter = Filters::new(HashMap::from([(filter_id, prefilter)]));
```
kespinola, line 37: I also want to create 1 subscription for all the filters instead of 1 per filter.

crates/yellowstone-grpc-source/src/simple_connector.rs
grooviegermanikus: done

PR #177, opened 2025-12-26
Fix BlockPrefilter::merge include flags to follow union semantics

RedaRahmani:
## Summary
While reading through the prefilter merge path, I noticed `BlockPrefilter::merge` was treating `include_*` flags as “last write wins” (`self = other`). That makes the final result depend on merge order and can accidentally drop requested block content (e.g., one parser sets `include_transactions=true`, another merges in later with `false` and turns it off).

`Prefilter::merge` is described as a union, so for “include” flags the expected behavior is: if any parser asks for it, we keep it enabled.
## Change
- Merge `include_accounts`, `include_transactions`, and `include_entries` using OR semantics so they stay monotonic (once true, always true).

- Keep `accounts_include` as a set-union (unchanged).
## Tests
Added unit tests to pin down the behavior:

- `true` is preserved across merges (`true + false => true`)

- merge remains stable across common merge patterns (and through the higher-level Prefilter merge/collect path if applicable)

## Notes
This keeps behavior consistent with the “union of prefilters” intent and removes an order-dependent footgun.

## FYI (unrelated to this change):
While validating locally, I noticed a couple of existing issues on main:

- `cargo clippy --all-targets --tests -- -D warnings` currently fails in `crates/spl-token-extensions-parser` due to `clippy::get_first` (`ix_data.get(0)` → suggested `ix_data.first()`).

- `cargo test` currently fails in `yellowstone-vixen-proc-macro` (`render::tests::test_generate_vixen_parser` panics with “Generated code should be valid Rust…”).

I didn’t include fixes here to keep this PR focused, but I’m happy to open follow-up PRs for these if you’d like.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -934,10 +934,7 @@ mod tests {
         let mut b = b_orig.clone();
         b.merge(a_orig.clone());
 
-        assert_eq!(
-            a, b,
-            "merge(A, B) should equal merge(B, A) (commutativity)"
-        );
+        assert_eq!(a, b, "merge(A, B) should equal merge(B, A) (commutativity)");
     }
 
     #[test]


RedaRahmani:
Thanks for the review and for merging, @kespinola !

PR #218, opened 2026-03-09
add handler methods "return from cpi frame" and others

grooviegermanikus:
### Problem
Some complex transactions (like Jupiter Swaps) do not emit a single CPI Log to wrap up an instruction but requires parser to collect multiple CPI Logs:
See this [Jupiter Swap](https://solscan.io/tx/31W4xWjSbYfNif2rdvZJNgk78zJfB1eRNh4MWA6fqoaqmLERtnCqZHeDromi6F8unXzmGFXbCniV7p3WCnJcxJio)  which has 3 `SwapEvent` events (at 6.2, 6.4, 6.6).

note: Jupiter uses tow styles (N `SwapEvent` events OR 1 `SwapEvents` event); for latter see this [Jupiter Swap](https://solscan.io/tx/2AspWZBmRkbmbMRxtbEwYXb6j3Pkz1dmvzpw1pK7CZnBmJhRFEv1JaLV2bYB13Hyzed4gRcbue98LhGo62xCsmQ4) with one CPI log at 3.6

### Discussion
* naming of methods might need to be improved

### Example
from `filtered-pipeline`:
```
--- starttx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
   >>> 2 ENTER
     > 2.1 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
     > 2.3 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
     > 2.4 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
   <<< 2 RETURN
   > 4 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
   >>> 5 ENTER
     >>> 5.1 ENTER
       > 5.1.1 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
       > 5.1.2 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
     <<< 5.1 RETURN
   <<< 5 RETURN
   > 6 tx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
=== endtx 2LSjh671hqDY8WUZDkveVcL5s2MCHWKCj3eLfmCoN4bNivVpUzCHEJJyKg2yDpVpuxW4v7WSHFZgL1eR67XXwm8p
```


crates/core/src/instruction.rs
```diff
@@ -436,43 +436,105 @@ impl InstructionUpdate {
...
+                let invoking_node_is_leaf = invoking_node.is_leaf();
+                let invoking_node = invoking_node.debug_node();
+                // let _last_path = last.1.clone(); // same as popped
+                // let sig = last.3.clone();
+                // let Some(ix) = last.0.next() else {
+                let Some(ix) = last.next() else {
                     let _ = d.pop_back().unwrap_or_else(|| unreachable!());
-                    continue;
+
+                    if !invoking_node_is_leaf {
```
grooviegermanikus: this is the actual magic

crates/runtime/src/instruction.rs
```diff
@@ -53,6 +87,11 @@ impl InstructionPipeline {
             }
         }
 
+        for pipe in &*self.0 {
+            pipe.handle_lifecycle(txn, instruction_shared, &LifecycleEvent::TxEnd)
```
kino-andrew, line 94: What strategy do we want to apply to lifecycle ? Right now I think we could have incomplete lifecycle because of the fail fast we have on pipeline, for example:

1. TxStart
2. PhysicalNode(3)
3. CpiEnter { caller_cpi_path: 3 }
4. PhysicalNode(3.1) -> handler returns Err
5. return immediately

We would miss 
5. CpiReturn { caller_cpi_path: 3 }
6. TxEnd

grooviegermanikus: good question. my 2ct: fail fast should be okey. if handler need to be smarter the implementor needs to catch and deal with error conditions.

kino-andrew: Ok, then we should be explicit and comment that the lifecycle can be broken and shouldn't be relied on as a always complete, what do you think ?

grooviegermanikus: yes defo. do you want to change that?

kino-andrew: Yes, adding a comment + a test highlighting this would be great addition !

Example of comment:

```
/// / ! \ Lifecycle delivery is fail-fast. / ! \
/// If a lifecycle handler returns an error, dispatch stops immediately and
/// subsequent lifecycle events for the current transaction are not guaranteed
/// to be emitted. Consumers must not rely on receiving a complete, balanced
/// lifecycle sequence (for example `CpiEnter` without `CpiReturn`, or `TxStart`
/// without `TxEnd`).
```

grooviegermanikus: doing

grooviegermanikus: ah wait, the HandlerResult gets lost along the way .. should I change that ?

kino-andrew: Yeah atm it just swallows the errors true, we should propagate the error

grooviegermanikus: pushed fix for error propagate plus some tests

crates/runtime/src/handler.rs
```diff
@@ -197,6 +276,16 @@ pub trait DynPipeline<T>: std::fmt::Debug + ParserId + GetPrefilter {
...
         value: &'h T,
     ) -> Pin<Box<dyn Future<Output = Result<(), PipelineErrors>> + Send + 'h>>;
+
+    /// Optional callback for lifecycle events (tx start/end, CPI enter/return).
+    fn handle_lifecycle<'h>(
+        &'h self,
+        _txn: &'h TransactionUpdate,
+        _instruction_shared: &'h InstructionShared,
+        _event: &'h LifecycleEvent<'h>,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + 'h>> {
```
kino-andrew: We should have `Pin<Box<dyn Future<Output = Result<(), PipelineErrors>> + Send + 'h>>` here no ? 

Same for Pipeline impl

```
fn handle_lifecycle<'h>(
    &'h self,
    txn: &'h TransactionUpdate,
    instruction_shared: &'h InstructionShared,
    event: &'h LifecycleEvent<'h>,
) -> Pin<Box<dyn Future<Output = Result<(), PipelineErrors>> + Send + 'h>> {
    Box::pin(Pipeline::handle_lifecycle(self, txn, instruction_shared, event))
}
```

and in FIlterPipeline

```
fn handle_lifecycle<'h>(
    &'h self,
    txn: &'h TransactionUpdate,
    instruction_shared: &'h InstructionShared,
    event: &'h LifecycleEvent<'h>,
) -> Pin<Box<dyn Future<Output = Result<(), PipelineErrors>> + Send + 'h>> {
    Box::pin(FilterPipeline::handle_lifecycle(self, txn, instruction_shared, event))
}
```

So that in https://github.com/rpcpool/yellowstone-vixen/pull/218/changes#diff-ae2c13320aa6399d65c4529d3355ed440474b4f53943ac21723f858b962e991d we will treat lifecycle callbacks the same as handlers errors, what do you think ?

crates/runtime/src/instruction.rs
```diff
@@ -48,11 +82,19 @@ impl InstructionPipeline {
                 match res {
                     Ok(()) => (),
                     Err(PipelineErrors::AlreadyHandled(h)) => h.as_unit(),
-                    Err(e) => err = Some(e.handle::<InstructionUpdate>(&pipe.id())),
+                    Err(e) => {
+                        err = Some(e.handle::<InstructionUpdate>(&pipe.id()));
+                        break 'outer;
```
kino-andrew, line 87: Ok so this is changing InstructionPipeline to become fail-fast as SingleInstructionPipeline is, when atm InstructionPipeline just logs errors and continues, which implies a big handling shift. 

Looking through this, I found that we have a comment which is outdated since the runtime implementation now wraps each registered instruction parser in SingleInstructionPipeline (new way to match runtime model).

The comment:
https://github.com/rpcpool/yellowstone-vixen/blob/23148a5bc1cf463b39e508622226f6027955d9e3/crates/runtime/src/builder.rs#L149-L158

The builder wrapping SingleInstructionPipeline https://github.com/rpcpool/yellowstone-vixen/blob/23148a5bc1cf463b39e508622226f6027955d9e3/crates/runtime/src/builder.rs#L209-L219

Then the question is: "fail-fast" or "log and continue" ? For both or we keep the different behaviour?

Summoning @kespinola about this architecture design, maybe i'm wrong here 🥷

kespinola (changes requested):
@grooviegermanikus do you require these callbacks if you work with a TransactionUpdate? In your handler you get the full transaction and apply the parsers to the instructions as you need them so you keep the full context of the transaction without callbacks sprinkle?

kespinola (commented):
If we do this I'd like to set the runtime to have a different core like we do with the sources.

`yellowstone_vixen::Runtime<YourCustomCore, GrpcSource>::builder()`

grooviegermanikus:
closing the as this is not in scope for vixen; as workaround either use "programEvents" feature or extract the look in `SingleInstructionPipeline` and feed it from `TransactionUpdate` (not tested)

PR #243, opened 2026-04-02
feat: Backfill data report q1

kino-andrew:
Runtime:
- fix: end of stream was not draining existing channels before ending

Kafka sink:
- feat: add tx_index in transaction message headers (indicative value)

Jetstreamer:
- fix: use end slot with same semantics as firehose
- feat: add a channel to handle the possible leader skipped

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -311,7 +311,7 @@ impl<S: SourceTrait> Runtime<S> {
         match stop_ty {
             StopType::Signal(Ok(Some(s))) => {
                 tracing::warn!("{s:?} received, shutting down...");
-                Self::stop_buffer(buffer).await;
+                Self::force_stop_buffer(buffer).await;
                 Ok(())
             },
             StopType::Signal(Ok(None)) => Err(std::io::Error::new(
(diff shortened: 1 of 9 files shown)


crates/runtime/src/lib.rs
```diff
@@ -328,8 +328,9 @@ impl<S: SourceTrait> Runtime<S> {
                     Ok(())
                 },
                 SourceExitStatus::Completed => {
-                    tracing::info!("Source completed successfully");
-                    Ok(())
+                    tracing::info!("Source completed successfully; draining runtime buffer");
+                    should_stop_buffer = false;
+                    buffer.wait_for_stop().await
```
kespinola, line 332: Can you explain the overriding of the should_stop_buffer? Isn't the wait_for_stop already handling halting the execution of the block until the buffer is drained? basically not letting stop_buffer be reached unil the wait_for_stop has yielded?

kino-andrew: Yes it's correct but the issue there was that we shouldn't end up "stop_buffer" which actually force-stop the buffer and prevent drain error propagation.

Also renamed the helper to be more explicit that it's a force stop and not just a normal stop

kespinola (approved):

PR #249, opened 2026-04-29
core: add InstructionUpdate::direct_log_messages() (#186)

senzenn:
## Summary

Adds a CPI-aware iterator that yields only the log lines emitted while this instruction's program is at the top of the invocation stack — built on the existing `log_range` infrastructure from #236, instead of re-introducing a redundant `log_indices: Vec<usize>` field.

Also extracts a `LogLineKind` classifier and routes `assign_logs_recursive`, `find_invoke`, and `split_logs_by_outer_ix` through it. This fixes a preexisting bug where `Program log:` payloads containing `" invoke ["` or `" success"` substrings were misclassified as real invokes/closes, truncating `log_range`.

## Why this shape

Per @kespinola's review on #187: rather than add `log_indices: Vec<usize>` (duplicates the merged `log_range: Range<usize>`), this builds the requested CPI-aware view as a single additive method on `InstructionUpdate`. No struct field changes, no breaking API.

## Usage

```rust
// Outer Raydium ix that CPIs into the token program for transfers.
async fn handle(&self, ix: &InstructionUpdate) -> HandlerResult<()> {
    // Old: yields token-program "Instruction: Transfer" lines too.
    let _full = ix.log_messages();

    // New: yields only the lines Raydium emitted directly.
    for line in ix.direct_log_messages() {
        if let Some(payload) = line.strip_prefix("Program log: ray_log: ") {
            // parse Raydium's swap log...
        }
    }
    Ok(())
}
```
- Closes #186
- Supersedes #187

### DIFF
--- a/crates/core/src/instruction.rs
+++ b/crates/core/src/instruction.rs
@@ -248,6 +248,53 @@ impl InstructionUpdate {
     #[must_use]
     pub fn log_messages(&self) -> &[String] { &self.shared.log_messages[self.log_range.clone()] }
 
+    /// Returns the log messages emitted *directly* by this instruction's
+    /// program, excluding lines emitted while an inner CPI is on top of the
+    /// invocation stack.
+    ///
+    /// Unlike [`log_messages`](Self::log_messages), which yields every line
+    /// inside this instruction's `invoke`/`success` window (including nested
+    /// CPI logs), this iterator skips any line that occurs while a deeper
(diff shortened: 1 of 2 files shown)


senzenn:
@kino-andrew

kino-andrew:
Hey ! Thanks for the contrib, it's a good improvement. Can you run the fmt please ?

senzenn:
done :)

PR #259, opened 2026-05-19
feat: add flat index to ix because path being the call tree it's not usable for kafka headers

kino-andrew:


crates/kafka-sink/src/handler.rs
```diff
@@ -106,18 +106,18 @@ impl vixen::Handler<TransactionUpdate, TransactionUpdate> for BufferingHandler {
         };
 
         for ix_update in &instructions {
-            for ix in ix_update.visit_all() {
```
kespinola, line 109: Can vist_all method be deprecated now?

kino-andrew: No I don't think so, it fills a different purpose than visit_all_with_flat_indices and is also used in runtime handle()

kespinola: I see it used by with_flat-indices apologies.

kespinola (approved):

PR #260, opened 2026-05-19
fix single instruction pipeline dropping cpi ixs after parse error

mooncitydev:
hey team, its moondev. was looking at #184 and the spl token-2022 extension handler basically never firing for most ixs.

root cause is `SingleInstructionPipeline::handle` in `crates/runtime/src/instruction.rs`. when one ix fails to unpack (common in a txn full of cpi, regular spl token ixs mixed with extension ixs), it returned immediately. everything after that in `visit_all()` never ran. so you get almost no `MetadataPointer` and similar events even though they are in the txn.

`InstructionPipeline` already keeps iterating and only reports the error at the end. this pr makes the single-parser path do the same thing.

added a small unit test: first ix fails parse, second still hits the handler.

fixes #184

### DIFF
--- a/crates/runtime/Cargo.toml
+++ b/crates/runtime/Cargo.toml
@@ -25,6 +25,11 @@ yellowstone-grpc-client = { workspace = true }
 yellowstone-grpc-proto = { workspace = true }
 yellowstone-vixen-core = { workspace = true }
 
+[dev-dependencies]
+async-trait = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+yellowstone-grpc-proto = { workspace = true }
+
 [features]
 default = []
 opentelemetry = ["dep:opentelemetry"]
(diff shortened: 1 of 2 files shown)


kespinola (approved):

kespinola:
@mooncitydev checkout the CI runs. Thank you for your PR.

senzenn:
Thanks for the fix. The runtime behavior change looks reasonable, but this PR is not mergeable yet because local CI-equivalent checks fail.

Blocking issues:
- `cargo test --workspace` fails to compile: the added test parser uses `type Output = u8`, but `Parser::Output` must implement `prost::Message`.
- `cargo clippy --all-targets --tests --no-deps -- -Dwarnings` fails because `DynPipeline` is imported but unused.
- `cargo +nightly-2026-02-25 fmt --all -- --check` fails on import ordering in `crates/runtime/src/instruction.rs`.

Please fix those and rerun CI.

PR #263, opened 2026-06-05
jetstreamer-source: use firehose StatsTracking for progress logging

senzenn:
The handler kept its own `blocks_processed: AtomicU64` just to log progress every 10k blocks, while `firehose()` already tracks all of this and we were passing `stats_tracking: None`.

- new `stats_interval_slots` config field (default 10000, 0 disables)
- register a `StatsTracking` callback that logs slots/blocks/transactions/entries/leader-skipped + tps; the callback fires per worker thread so `thread_id` is included
- drop the atomic counter and its two logging sites
- add `--stats-interval-slots` to the jetstreamer example

Tested against the old-faithful archive (50 slots of epoch 885, interval 10):

```
Firehose progress thread_id=0 slots=11 blocks=11 transactions=12435 entries=9535 leader_skipped_slots=0 tps=2902
Firehose progress thread_id=2 slots=24 blocks=24 transactions=27338 entries=20563 leader_skipped_slots=0 tps=417
Firehose progress thread_id=1 slots=47 blocks=47 transactions=53267 entries=33296 leader_skipped_slots=0 tps=3578
```


examples/jetstreamer/src/main.rs
```diff
@@ -152,6 +152,10 @@ struct Opts {
     /// Archive URL
     #[arg(long, default_value = "https://api.old-faithful.net")]
     archive_url: String,
+
+    /// Emit a firehose progress stats line every N slots (0 disables)
+    #[arg(long, default_value = "10000")]
+    stats_interval_slots: u64,
```
kespinola, line 158:
```suggestion
    /// Emit a firehose progress stats line every N slots (0 disables)
    #[arg(long, default_value = "10000")]
    stats_interval_slots: Option<u64>,
```

crates/jetstreamer-source/src/lib.rs
```diff
@@ -656,6 +690,17 @@ impl JetstreamSource {
...
+        // Register upstream `StatsTracking` so periodic progress comes from the
+        // firehose's own aggregates. `stats_interval_slots == 0` disables it —
+        // this also avoids upstream's unguarded `slot % interval` (a `0`
+        // interval would divide by zero).
+        let stats_tracking = (config.stats_interval_slots != 0).then_some(
+            jetstreamer_firehose::firehose::StatsTracking {
+                on_stats: log_firehose_stats,
+                tracking_interval_slots: config.stats_interval_slots,
+            },
+        );
```
kespinola, line 702:
```suggestion
        let stats_tracking = config.stats_interval_slots.map(|tracking_interval_slots| {
            jetstreamer_firehose::firehose::StatsTracking {
                on_stats: log_firehose_stats,
                tracking_interval_slots,
            }
        });
```

If go as optional can use a map and the None to signal that logging should be disabled.

crates/jetstreamer-source/src/lib.rs
```diff
@@ -598,6 +589,49 @@ impl SourceTrait for JetstreamSource {
...
+/// Example output:
+///
+/// ```text, ignore
+/// INFO Firehose progress thread_id=2 slots=20000 blocks=19987 transactions=4821334 entries=20000 leader_skipped_slots=13 tps=152000
+/// ```
+fn log_firehose_stats(
+    thread_id: usize,
+    stats: jetstreamer_firehose::firehose::Stats,
+) -> futures_util::future::BoxFuture<'static, Result<(), SharedError>> {
+    async move {
```
kespinola, line 611: Do we have to return a future or can the function be made async and passed as the callback?

kespinola (changes requested):

crates/jetstreamer-source/src/lib.rs
senzenn: `.map` needs the field to be `Option<u64>` (it's `u64` right now), and the `!= 0` check is load-bearing — upstream does `slot % tracking_interval_slots` unguarded ([[firehose.rs:1575](https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/jetstreamer-firehose/src/firehose.rs#L1575)](https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/jetstreamer-firehose/src/firehose.rs#L1575)), so a 0 has to become `None` or it panics. A plain `.map` passes `Some(0)` straight through and brings the panic back.

Happy to move to `Option<u64>`, just keeping the guard:

```rust
let stats_tracking = config
    .stats_interval_slots
    .filter(|&n| n != 0)
    .map(|tracking_interval_slots| jetstreamer_firehose::firehose::StatsTracking {
        on_stats: log_firehose_stats,
        tracking_interval_slots,
    });
```

Want it opt-out (default `Some(10_000)`) or opt-in (default `None`)? I'll fix up the defaults and tests to match.

crates/jetstreamer-source/src/lib.rs
senzenn: Has to be a boxed future upstream's `Handler<Data>` trait requires `Fn(usize, Data) -> BoxFuture<'static, HandlerResult>`, so a bare `async fn` won't satisfy the bound. `.boxed()` erases the opaque future into the `BoxFuture` it expects.

Same shape as the existing `on_block`/`on_tx`/`on_entry` callbacks at lib.rs:663–681. Happy to inline it as a closure next to those if you'd prefer.

examples/jetstreamer/src/main.rs
senzenn: Keeping `u64`: the config expects `u64`, and `0` already disables tracking.

PR #265, opened 2026-06-18
fix: infinite loop rebuilding cpi tree when inner instruction stack_height is missing

mooncitydev:
hey, its moondev. was reading through the instruction building in `core` and i think theres a hang hiding in `attach_inner_instructions`.

when it rebuilds the cpi call tree from inner instruction stack heights it walks the list backwards. the relevant bit:

```rust
while i > 0 {
    let parent_idx = i - 1;
    let Some(height) = inner[parent_idx].1 else {
        // stack_height missing for old data
        continue;        // i never gets decremented
    };
    ...
    i -= 1;
}
```

if a parent instruction has no `stack_height`, the let-else does `continue` but never decrements `i`, so it just spins on the same index forever. one transaction like that pins a worker at 100% cpu, and since the buffer runs a bounded pool of these jobs, enough of them will stall the whole pipeline.

its not a theoretical input either. the function right next to this one (`derive_paths_from_stackheights`) already has a branch for missing stack heights ("old data"), so `None` is clearly expected, and any source that doesnt populate `stack_height` at all makes every inner group with 2+ ixs hit it (the first `parent_idx` it checks is `None` and it locks up).

fix is one line, decrement `i` before continuing so a missing height just leaves that ix flat and we move on, which is what the existing comment already says it intends to do. i added a regression test that builds a txn whose inner ixs all have `stack_height = None` and asserts `build_from_txn` returns (and keeps them flat) instead of hanging.

on testing: i pulled the exact loop into a standalone repro to be sure, current code never returns (had to guard it with a timeout thread), and with the fix it returns right away and still nests correctly when heights are present. heads up that i couldnt run the full `cargo test` on my windows box (no msvc/mingw linker handy here), so CI is the real check, but its tiny and self contained.

cheers

### DIFF
--- a/crates/core/src/instruction.rs
+++ b/crates/core/src/instruction.rs
@@ -852,6 +852,40 @@ mod tests {
         assert!(instructions[0].inner.iter().all(|i| i.inner.is_empty()));
     }
 
+    #[test]
+    fn build_from_txn_nests_present_stack_heights_when_one_is_missing() {
+        use super::Pubkey;
+
+        // Companion to the all-`None` case above: with a mix of present and
+        // missing `stack_height`s the reconstruction must still terminate, nest
+        // the instructions whose heights are known, and leave the height-less
+        // one flat. Heights here are [Some(2), Some(3), None]: the depth-3
+        // instruction is a child of the depth-2 one, and the instruction with no
+        // height stays a sibling at the top of the inner list.
+        let txn = transaction_with_mixed_inner_stack_heights();
+
+        let instructions =
+            InstructionUpdate::build_from_txn(&txn).expect("transaction should build");
+
+        assert_eq!(instructions.len(), 1);
+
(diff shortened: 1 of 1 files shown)


senzenn:
Confirmed the hang and verified the fix locally (tests + clippy pass) — i -= 1 is right. One ask before merge: the regression test only covers all-None heights. Could you add a mixed case (e.g. Some(2), Some(3), None)? That's where the nesting logic actually does work. Otherwise LGTM.

senzenn:
Reviewed and tested locally. The fix is focused and the regression test covers the missing `stack_height` infinite-loop case.

Passed locally:
- `cargo test --workspace`
- `cargo +nightly-2026-02-25 fmt --all -- --check`
- `cargo clippy --all-targets --tests --no-deps -- -Dwarnings`

Looks mergeable to me once CI/branch protection is satisfied.

PR #279, opened 2026-07-26
Rename to Shipstern

kespinola:
Kyle approved the wire break, so the protobuf packages are renamed here and this goes out as 0.8.0. Supersedes the revert in 8406cfa.

Renaming to Shipstern #280

## The break

`vixen.stream`, `vixen.parser.token`, `vixen.parser.bpf_loader` and `vixen.parser.token_extensions` are now `shipstern.*`.

Those strings are the wire contract. They set the gRPC method path, now `/shipstern.stream.ProgramStreams/Subscribe`, and the `Any` type URLs in `SubscribeUpdate.parsed`, now `type.googleapis.com/shipstern.parser.token.TokenAccount`.

Existing gRPC consumers have to regenerate and redeploy. There are two failure modes if they don't. The method path returns `UNIMPLEMENTED`, which is obvious. The `Any` type URLs stop matching without erroring, which isn't: a consumer that unpacks `SubscribeUpdate.parsed` by type URL just stops resolving updates, with nothing in the logs. The second one is the one to warn people about. No shim, and the old packages aren't served alongside the new ones.

Past the four package lines, this covers the 144 fully-qualified `shipstern.parser.token.*` cross-references in `token_extensions.proto`, the descriptor-set paths in `build.rs`, and the `include!` / `include_proto!` paths in `src/lib.rs`.

`wire_compat_tests` and `stream_wire_compat_tests` now pin `shipstern.*` and fail on any leftover `vixen.*` declaration, so a half-finished sweep trips CI instead of shipping. Kept the step pinning `--features parser,stream`.

## 0.8.0

Workspace version, the 22 internal path pins, and both lockfiles. `spl-pod` and `spl-token-group-interface` stay at `^0.7.0`; they just happen to share the number. `publish` stays false.

## Split out the program-events tests

The 5 snapshots and ~8,062 lines are off this branch now: 5 test files, 11 cfg arms, 5 snapshot files. They add program-events coverage to `tests/proc-macro` and aren't rename work. The snapshots declare per-program packages (`package perpetuals;`), so there's no vixen or shipstern naming in them to begin with. Kept on a local branch for a separate PR.

## Merged main

Three conflicts. `ci.yml` keeps both steps, the wire-compat one and #283's experimental-account-parser one. Both lockfiles regenerated instead of hand-merged.

#283's clippy step also came in pointing at `yellowstone-vixen-kafka-sink`, which doesn't exist here, so cargo rejected the feature spec before linting anything. That was the red clippy run, fixed in 37a5e10. It merged without a conflict, unlike the test job's equivalent step, so it arrived quietly.

`main` still has the same stale reference in its own clippy step. Harmless there because the package exists, but it'll break for whoever merges this.

## Verified locally

Against the pinned toolchain:

- `fmt --all -- --check`: clean
- `clippy --all-targets --tests --no-deps --features shipstern-kafka-sink/experimental-account-parser -- -Dwarnings`: clean
- `test --workspace --lib --tests`: 249 passed
- `test -p shipstern-proto --lib --features parser,stream`: 5 passed, including both flipped guards
- `test -p shipstern-kafka-sink --features experimental-account-parser --lib --tests`: 36 passed
- `tests/proc-macro-events`: 33 passed
- `cargo tree --frozen --offline`: passes in both workspaces, so the lockfiles are consistent

The generated descriptor sets carry `shipstern.parser.*` and no `vixen` strings.

## Needs you

- #280 still says the API doesn't change. It needs the consumer note now.
- Transfer and rename both need org admin.
- Don't publish before the transfer lands. Those URLs resolve only after the move and crate metadata is immutable. The old `yellowstone-vixen-*` crates stay frozen.
- #280 promises the old crates emit deprecation warnings for 180 days. Still nothing implements that.


senzenn:
Pushed 8406cfa, 4983947, b6977cb. One thing needs your call, rest is just FYI.

The rename changed the protobuf packages (vixen.stream, vixen.parser.*) to shipstern.*. Those are the gRPC wire contract, not just names they set the method path and the Any type URLs, so renaming them breaks deployed clients (UNIMPLEMENTED, plus type URLs that silently stop matching). I reverted them for now, so proto/ and build.rs match main; only the crate/module stay shipstern. If you want the break, say so and I'll put it back but then it's 0.8.0 and #280 needs a consumer note.

Also fixed: repo URLs now point at solana-rpc/shipstern (rpcpool/shipstern would never resolve); kept the 43 CHANGELOG links on the old repo so the redirect carries them; added wire-compat tests since nothing covered the package strings before.

Needs you: transfer + rename need org admin; don't publish before the transfer lands; ~19 crates is a couple hours under crates.io rate limits; #280 still promises 180-day deprecation warnings nobody's built. Branch also still carries the program-events tests (~8k lines) want me to split those out?

Also patched the #280 migration script: \b is GNU-only so the bare rules no-op on macOS, and include_vixen_parser needed the ! dropped to catch the import. Tested both.

senzenn:
Heads up that #287 touches `crates/jetstreamer-source/src/lib.rs` and will conflict with this branch.

As things stand this PR merges clean into main. With #287 in, a 3-way merge gives 2 conflict hunks in that file, and the new code there adds 3 `VixenStreamHandler` and 4 `yellowstone_vixen_core` references the rename here has not seen.

Simplest order is this PR first, then I rebase #287 on top. That also gets the `deny.toml` ignores for RUSTSEC-2026-0247/0248/0251 in, which are currently failing the advisories check on every PR against main.

PR #287, opened 2026-08-15
fix(jetstreamer-source): deliver block meta and slot updates to their pipelines

senzenn:
## Problem

`block_meta` and `slot` prefilters were bucketed into the block match list, so their filter IDs were attached to `UpdateOneof::Block`. The runtime dispatches by variant (`crates/runtime/src/buffer.rs:128/141/154`), so those IDs matched nothing in `pipelines.block` and were dropped with only a `trace!` line (`handler.rs:325`).

Block-meta and slot parsers therefore received nothing while the run reported success. `yellowstone-vixen-block-coordinator`, which keys on `BlockMeta` (`coordinator.rs:145`, `source.rs:275`), could not be driven from the historical-replay path at all.

## Fix

Bucket filter IDs by the `UpdateOneof` variant that actually consumes them, and emit `SubscribeUpdateBlockMeta` and `SubscribeUpdateSlot` alongside `Block`. Both messages are built from fields already carried on `BlockData::Block`, so no additional data is needed and no public API changes.

A `send` helper collapses four near-identical `SubscribeUpdate` construction sites into one. The entry buffer now drains on every block rather than only on the emitting path, so a slot no filter cares about cannot pin its entries.

Replayed slots report `SLOT_FINALIZED`. Old Faithful archives hold only finalized history and cannot reproduce the processed/confirmed transitions that `SlotPrefilter::filter_by_commitment = false` asks for.

## Verification

Live A/B against mainnet epoch 885, slots 382320000..382320005, running the same instrumented example binary against both versions of the source with `block_meta` and `slot` pipelines registered.

Before (unpatched):
* `block_filters=2`
* BlockMeta handler hits: 0
* Slot handler hits: 0

After (this branch):
* `block_filters=0 block_meta_filters=1 slot_filters=1`
* BlockMeta handler hits: 5
* Slot handler hits: 5

Both runs: 5 slots, 5 blocks, 5399 transactions, 3499 SPL Token instructions parsed, exit 0, 0 firehose errors. The unpatched run processed every block and logged "replay completed successfully" while delivering nothing to either pipeline. The identical instruction count on both sides shows the transaction path is untouched.

Delivered data checks out against the archive: parent chain contiguous (382319999 through 382320004), slot status 2 (`SlotFinalized`), per-block rewards 852 to 4043, and per-block transaction counts summing to 5399, matching the firehose summary total.

Sequential mode re-checked separately (`--sequential`, 3 slots, 36s, 0 errors); the 256 MiB bounded window still applies and does not hang.

Adds three unit tests: block-meta only, slot only, and all three variants emitted for one block.

## Gates

* `cargo +nightly fmt --all -- --check`
* `cargo clippy --workspace --all-targets -- -D warnings`
* `cargo test --workspace` (253 passed, 0 failed)

## Not addressed here

Two related issues in this file are left for separate PRs: block updates still carry empty `transactions`/`accounts` even when `BlockPrefilter` requests them, and the per-slot entry buffer can retain entries for a slot whose block upstream suppresses on a retry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](https://semver.org/).
 ### Changed
 
 - Bumped `yellowstone-grpc-proto` from `12.4` to `12.6`, picking up the Transaction V1 (SIMD-0385) `Message.config` field and the `cuckoo_account_include` transaction filter. The new filter is not exposed through `TransactionFilter` and is always sent as `None` ([#304](https://github.com/solana-rpc/shipstern/pull/304) by @ultrasilicon).
+- `yellowstone-vixen-jetstream-source`: a parser registered on the `block` pipeline whose prefilter sets `block_meta` or `slot` but none of `include_transactions` / `include_accounts` / `include_entries` no longer receives a `Block` update. Previously any `block_meta`/`slot` prefilter forced its filter ID into the block match list, so such a parser received a header-only `Block`: `transactions` and `accounts` were always empty. Use a `block_meta` pipeline instead.
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


senzenn:
Note on the red check: the failure is the advisories step inside the `fmt` job, not rustfmt. It trips on three unmaintained crates that come in through solana-runtime 3.1.4: RUSTSEC-2026-0247, 0248 and 0251.

It is not from this PR. This changes two files and no dependency manifests, so the lock file is identical to main. rustfmt itself passes on the pinned nightly-2026-02-25.

#279 already adds those three ignores to `deny.toml`, so merging that clears it.

PR #295, opened 2026-08-25
feat(proc-macro): expose discriminator constants on generated types

### LINKED ISSUE #294: Expose discriminator constants on `include_vixen_parser!` generated types
The proc macro - include_vixen_parser emits no discriminator const or accessor for accounts or instructions. Bytes appear only as unnamed literals inside comparisons, in both `render/account_parser.rs` and `render/instruction_parser.rs`:

```rust
if slice == &[#(#discriminator_bytes),*] { ... }
```

They don't reach the parsed value either — deserialization starts at `data[end..]` in both paths.

Instructions are partly covered: `InstructionResolver::resolve` gets `data: &[u8]`. Accounts have nothing equivalent.

Solution : 
    Expose ix and acc discriminators either as a const fn or a constant

```rust
impl BondingCurve { pub const DISCRIMINATOR: &'static [u8] = &[...]; }
```

senzenn:
Closes #294.

Discriminator bytes only lived as literals inside the comparison checks and got thrown away after codegen, so nothing downstream could read them. This exposes them on the generated types.

```rust
impl BondingCurve {
    pub const DISCRIMINATOR: &'static [u8] = &[23, 183, 248, 55, 96, 216, 172, 96];
    pub const DISCRIMINATOR_OFFSET: usize = 0;
}

impl Instructions {
    pub const BUY_DISCRIMINATOR: &'static [u8] = &[102, 6, 61, 18, 1, 218, 235, 234];
    pub const BUY_DISCRIMINATOR_OFFSET: usize = 0;
}
```

The pair is a memcmp predicate: these bytes sit at that offset on the wire and they are what the parser matches, so they drop straight into `Memcmp::new_base58_encoded(offset, bytes)` or the geyser account filter.

They are **not** a payload boundary. SPL Governance deserializes from `data[..]`, so byte 0 is re-read as the first body field: `RealmV2::DISCRIMINATOR` is `&[16]`, which is also its `account_type` tag. Decode with `try_unpack`. No `DISCRIMINATOR_LEN` on purpose, since `OFFSET + LEN` math is the wrong move for that case.

Additive: `instruction_parser.rs` is +120/-0, the `account_parser.rs` `quote!` bodies are unchanged, and no match, filter or deserialize logic is touched.

### Skipped cases

No const where the parser could not honor it: size-only discriminators, zero-length ones, fixed-size fields whose declared width does not match the decoded bytes, and instruction names that collide after case folding. That last one matters because `ä` and `Ä` are distinct instructions that both fold to `Ä_DISCRIMINATOR`, so emitting would break an IDL that compiles today.

### Tests

Absence is compiler-checked (E0599). Covered: Anchor field discs (`dynamic_bonding_curve`), numeric consts with the re-read behavior asserted (`spl_governance`), a byte const at offset 8 behind a prefix so the `offset + len` path runs (`constant_bytes_account`), all 29 pump_fun instruction consts driven through `resolve_instruction_default`, and the `Buy` const checked against a real mainnet tx. New fixtures cover the four skip cases.

### Found while in here (not fixed)

- Width-mismatch matcher bug, same cause in two spots. This PR suppresses the const, matcher untouched. #296
- `SHOUTY`/`parse_*` name collision on ASCII case-only instruction names. #297

### DIFF
--- a/crates/proc-macro/src/render/account_parser.rs
+++ b/crates/proc-macro/src/render/account_parser.rs
@@ -22,6 +22,40 @@ fn decode_discriminator_bytes(bytes: &codama_nodes::BytesValueNode) -> Vec<u8> {
 }
 
 ///
+/// Build the discriminator constants for a generated account type.
+///
+/// The pair is a memcmp predicate: `DISCRIMINATOR` is exactly what the parser
+/// compares at `DISCRIMINATOR_OFFSET`. It is not a payload boundary. A numeric
+/// discriminator is re-read as the first field of the account body, so callers
+/// must decode through `try_unpack` rather than slicing past it.
+///
+/// Returns `None` for an empty discriminator, which would match every account.
+///
+fn discriminator_consts(
+    ident: &proc_macro2::Ident,
+    bytes: &[u8],
+    offset: usize,
+) -> Option<TokenStream> {
+    if bytes.is_empty() {
+        return None;
+    }
+
+    Some(quote! {
+        impl #ident {
+            /// Discriminator bytes that identify this account type on the wire.
+            ///
+            /// Pairs with [`Self::DISCRIMINATOR_OFFSET`] as a memcmp predicate:
+            /// these bytes appear at that offset in the raw account data. It is
+            /// not a payload boundary. Decode with `try_unpack`.
+            pub const DISCRIMINATOR: &'static [u8] = &[#(#bytes),*];
+
+            /// Byte offset at which [`Self::DISCRIMINATOR`] begins in the account data.
+            pub const DISCRIMINATOR_OFFSET: usize = #offset;
+        }
+    })
+}
+
+///
 /// Build the *account parser* for a program.
 ///
 /// Generates a wrapper struct with a non-Option oneof field and a manual
@@ -108,7 +142,7 @@ pub fn account_parser(
         }
     });
 
-    let account_matches = accounts.iter().filter_map(|account| {
+    let account_entries = accounts.iter().filter_map(|account| {
         let discriminator = match account.discriminators.first() {
             Some(d) => d,
             None => {
@@ -133,7 +167,7 @@ pub fn account_parser(
 
                         let value_u8 = value as u8;
 
-                        quote! {
+                        let arm = quote! {
                             if let Some(discriminator) = data.get(#offset) {
                                 if *discriminator == #value_u8 {
                                     match <#account_ident as ::borsh::BorshDeserialize>::deserialize(&mut &data[..]) {
@@ -146,7 +180,9 @@ pub fn account_parser(
                                     }
                                 }
                             }
-                        }
+                        };
+
+                        (arm, discriminator_consts(&account_ident, &[value_u8], offset))
                     },
 
                     // Byte discriminators are a prefix and are not part of the account struct.
@@ -154,7 +190,7 @@ pub fn account_parser(
                         let discriminator = decode_discriminator_bytes(bytes);
                         let end = offset + discriminator.len();
 
-                        quote! {
+                        let arm = quote! {
                             if let Some(slice) = data.get(#offset..#end) {
                                 if slice == &[#(#discriminator),*] {
                                     match <#account_ident as ::borsh::BorshDeserialize>::deserialize(&mut &data[#end..]) {
@@ -167,7 +203,9 @@ pub fn account_parser(
                                     }
                                 }
                             }
-                        }
+                        };
+
+                        (arm, discriminator_consts(&account_ident, &discriminator, offset))
                     },
 
                     _ => return None,
@@ -224,7 +262,7 @@ pub fn account_parser(
                     quote! { slice == &[#(#discriminator),*] }
                 };
 
-                quote! {
+                let arm = quote! {
                     if let Some(slice) = data.get(#offset..#end) {
                         if #disc_check {
                             match <#account_ident as ::borsh::BorshDeserialize>::deserialize(&mut &data[#end..]) {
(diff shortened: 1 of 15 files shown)


tests/proc-macro/tests/pump_fun.rs
```diff
@@ -119,3 +119,267 @@
...
+}
+
+///
+/// Tie the constant to real on-wire bytes, not just to the matcher.
+///
+/// For the instruction the matcher resolves as `Buy`, the bytes actually present
+/// in the mainnet transaction at `BUY_DISCRIMINATOR_OFFSET` must equal
+/// `BUY_DISCRIMINATOR`.
+///
+#[tokio::test]
```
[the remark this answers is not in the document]
senzenn: Fixed in 4a9d02b. Both sinks the query traced now drop the value:

- `tests/proc-macro/tests/pump_fun.rs`: `panic!("expected an instruction fixture, got {other:?}")` becomes `panic!("expected an instruction fixture")`
- `crates/mock/src/lib.rs`: `panic!("parse error: {e:?}")` in `run_ix_parse!` becomes `panic!("parse error")`

Both bindings become `_` so the unused-variable lint stays clean.

Alert 49 now reports `state=fixed` on `refs/pull/295/head`, and the CodeQL check passes on the current head. This comment is from c9db7bb and predates the fix.

crates/proc-macro/src/render/account_parser.rs
```diff
@@ -21,6 +21,40 @@ fn decode_discriminator_bytes(bytes: &codama_nodes::BytesValueNode) -> Vec<u8> {
...
+        impl #ident {
+            /// Discriminator bytes that identify this account type on the wire.
+            ///
+            /// Pairs with [`Self::DISCRIMINATOR_OFFSET`] as a memcmp predicate:
+            /// these bytes appear at that offset in the raw account data. It is
+            /// not a payload boundary. Decode with `try_unpack`.
+            pub const DISCRIMINATOR: &'static [u8] = &[#(#bytes),*];
+
+            /// Byte offset at which [`Self::DISCRIMINATOR`] begins in the account data.
+            pub const DISCRIMINATOR_OFFSET: usize = #offset;
```
Nagaprasadvr, line 53: isnt this always 0 , ig we can remove this ?

tests/idls/constant_bytes_account.json
```diff
@@ -42,6 +42,41 @@
...
+                "format": "u8",
+                "endian": "le"
+              }
+            }
+          ]
+        },
+        "discriminators": [
+          {
+            "kind": "constantDiscriminatorNode",
+            "offset": 8,
```
Nagaprasadvr, line 66: this test should fail as offset should always be 0

crates/proc-macro/src/render/account_parser.rs
senzenn: It's 0 in our current fixtures, but it's `node.offset` off the codama node a required field with no serde default, so a non-zero one can show up any time. And main's matcher already slices `data.get(offset..end)` (account_parser.rs:158, :228), so non-zero offsets are already supported, not something this PR adds.

`Memcmp::new_base58_encoded(offset, bytes)` needs both. Drop the offset const and callers hardcode 0, which silently mis-filters the first non-zero IDL.

tests/idls/constant_bytes_account.json
senzenn: It's `node.offset` from the codama node, not a literal. All our IDLs are 0 today, but main's matcher already slices `data.get(offset..end)` (account_parser.rs ~158–228), so non-zero offsets are already handled. `Memcmp::new_base58_encoded` takes both args drop the const and callers assume 0 and mis-filter silently on the first non-zero IDL. That's what `round_trips_a_discriminator_at_a_nonzero_offset` guards.

crates/proc-macro/src/render/account_parser.rs
Nagaprasadvr: Okay if that is the case maybe we can keep it @kespinola wty?

tests/proc-macro/tests/pump_fun.rs
```diff
@@ -119,3 +119,267 @@ fn check_json_serialization() {
...
+/// decode paths. Driving `resolve_instruction_default` with a buffer built from
+/// each constant proves they agree: a mismatch surfaces as
+/// `DiscriminatorNotFound`. Argument deserialization is expected to fail on
+/// these synthetic buffers; only the discriminator verdict is asserted.
+///
+#[test]
+fn every_instruction_const_is_recognised_by_the_matcher() {
+    let all: &[(&[u8], usize, &str)] = &[
+        (
+            pump_fun::Instructions::ADMIN_SET_CREATOR_DISCRIMINATOR,
```
Nagaprasadvr, line 164: cant we scope these discs to its respective ixs ? like AdminSetCreator::DISCRIMINATOR

senzenn: yaya good call on type-scoped constants. AdminSetCreator::DISCRIMINATOR works directly and it drops the const_name_counts fold (a and A collide today). Worth doing before this lands, since switching later breaks anyone matching Instructions::FOO_DISCRIMINATOR.

One check: does the collision in build_instruction_messages only hit <IxName>Args, never a bare <IxName>?

Nagaprasadvr: Can u elaborate your question i didn't get it

senzenn: I meant: does that collision skip ever hit the bare <IxName> wrapper, or only <IxName>Args?

senzenn: Only `Args`. The instruction type like `AdminSetCreator` is always generated — the skip only ever drops the `Args` struct. Fixtures show no clashes, so `AdminSetCreator::DISCRIMINATOR` always has a type to sit on.

Only edge case is #297 (names that differ just by upper/lowercase), but that already fails to compile on its own, so this change adds nothing new.

crates/proc-macro/src/render/account_parser.rs
kespinola: Its in the codama spec to explicitly define the discriminator offset then lets support it in shipstern.

kespinola (approved):

Nagaprasadvr (approved):
Gtg

PR #302, opened 2026-09-01
feat(runtime): update gRPC filters on a live subscription

### LINKED ISSUE #286: Expose the  SubscribeRequest `Sink` when connecting to yellowstone grpc source
### Context

`yellowstone-grpc-client` returns both a stream and a sink when subscribing.
The sink accepts additional `SubscribeRequest` messages after the initial
subscription, which the server applies to the existing connection — this is how
filters are meant to be changed at runtime.

### Current behavior

Vixen consumes the stream but does not surface the sink to callers. Once a
`SubscribeRequest` is issued at startup, the filter set is effectively frozen
for the lifetime of the connection.

### Proposal

Expose the sink (or a thin wrapper over it) so downstream consumers can send
updated `SubscribeRequest`s to an active subscription.

### Motivation

Clients whose filter set is derived from dynamic configuration — accounts,
program IDs, or transaction filters that change while the process is running —
currently have to tear down and re-establish the gRPC connection to pick up
changes. That means dropped messages during the reconnect window, re-doing the
initial handshake, and a gap in coverage that is awkward to reason about for
anything monitoring-adjacent. Sending an updated request over the existing sink
avoids all of it.

source - https://github.com/rpcpool/yellowstone-vixen/blob/main/crates/yellowstone-grpc-source/src/lib.rs#L202

### Implementation

A new trait method can be implemented to return the sink instead of breaking the current implementation

senzenn:
`yellowstone-grpc-client` returns a sink alongside the update stream, and the
source drops it:

```rust
// crates/yellowstone-grpc-source/src/lib.rs:202 on main
let (_sub_tx, stream) = client
    .subscribe_with_request(Some(subscribe_request))
```

So the filter set is frozen for the life of the connection. A process whose
filters come from config that changes at runtime has to tear down the
connection and resubscribe, which drops everything in the gap.

## What this is not

It does not hand out the sink, which is what #286 proposed. The channel carries
`Filters` keyed by parser ID and the source builds the `SubscribeRequest`
itself. A raw request would skip the commitment fill-in and let the subscription
drift from the registered parsers, and it would make the runtime gRPC-shaped
when sources are meant to be generic. That trade is discussed on #286 and in the
source README.

## API

```rust
let runtime = Runtime::<YellowstoneGrpcSource>::builder()
    .account(Pipeline::new(TokenProgramAccParser, [Handler]))
    .try_build(config)?;

// `run` takes the runtime by value, so take the handle first.
let handle = runtime.handle();
tokio::spawn(runtime.run_async());

let extra = Prefilter::builder().account_owners([new_mint]).build()?;
handle.update_filters(|filters| filters.merge(TokenProgramAccParser.id(), extra))?;
handle.reset_filters()?;
```

`RuntimeHandle` is cheap to clone, every method is sync, and `Runtime::handle`
only exists for sources implementing `FilterUpdateSource`, so taking one from a
source that cannot change its subscription is a compile error.

Includes #305 and #307, both merged into this branch.

## Semantics worth knowing before review

- Every send replaces the whole subscription, so send the complete set.
- `Ok(())` means the set reached the source, not that the server applied it.
  There is no ack. The source also awaits the runtime buffer inside the same
  loop that watches for updates, so while that buffer is full the request has
  not reached the server either.
- `filters()` is the last set published and leads what the server is serving by
  an unbounded amount when the sink is between connections.
- Keys are parser IDs, except instruction parsers. The builder bundles all of
  them behind one `InstructionPipeline`, so the set holds a single entry under
  `InstructionPipeline::ID` and naming an individual instruction parser is
  refused as unknown.
- `UnknownParser` checks the name, not the prefilter kind. A prefilter whose
  kind does not match the pipeline still reaches the wire and is discarded.
- A set the server refuses comes back on the stream, not the sink. A terminal
  status code ends the run. A recoverable one does not: the sink records the
  request into its reconnect state before the server sees it, so the client
  resubscribes the same refused set and repeats. `ResourceExhausted` is
  recoverable and is what a "too many filters" rejection can carry.
- `from_slot` applies to the initial subscribe only. On an update geyser reads
  it as a replay request, and richat rejects it when the set contains blocks.

## Evidence

Live against a mainnet endpoint, with the SPL token account parser registered so
its prefilter is every token account. Each number is distinct accounts seen over
the window, after the queued backlog drains:

| step | distinct accounts |
| --- | --- |
| registered prefilter, 8s | 3185 |
| one account inserted through the handle before `run`, 20s | 1 |
| second account merged into the live subscription, 20s | 2 |
| `reset_filters`, 8s | 4170 |

The second row is the one that matters. It is also the regression this branch
fixes: the initial subscribe used to be seeded from the registered set, so a
caller narrowing before `run` got the wide set first and a second request behind
it, and `from_slot` replayed against the set they thought they had replaced.

Unknown parser IDs were refused without reaching the wire. No reconnect and no
stream error across four subscription changes on one connection.

CI gates all pass, including `cargo deny check advisories bans` and the
`test-events` workspace.

## Ordering

#308 is stacked on this branch and should land here before this merges, or
`SourceTrait` changes twice on the way to `main` and out-of-tree sources migrate
twice. The rebase is clean apart from two breaks git merges without a marker,
`try_build` now needing `S: FromConfig` and `default_test_config` being gone,
both in one test helper.

Closes #286.


crates/runtime/src/runtime_tests.rs
```diff
@@ -373,6 +382,131 @@ impl Handler<SlotUpdate, SlotUpdate> for SlowSlotHandler {
...
+#[derive(Debug)]
+struct MockFilterUpdateSource;
+
+#[async_trait]
+impl SourceTrait for MockFilterUpdateSource {
+    type Config = NullConfig;
+
+    fn new(_: NullConfig, _: Filters) -> Self { Self }
+
+    fn supports_filter_updates() -> bool { true }
```
Nagaprasadvr: i am wondering do we need this

crates/runtime/src/builder.rs
```diff
@@ -12,6 +13,16 @@ use crate::{
...
 };
 
+/// Depth of the filter update channel handed to callers by
+/// [`Runtime::filter_updates`]. Updates are rare, so a shallow queue is enough
+/// to keep a caller from blocking on a short burst.
+///
+/// Queued sets are forwarded in order and the server ends on the newest. The
+/// exception is a set the sink rejected: the source holds that one and a newer
+/// arrival replaces it rather than queueing behind it, so an intermediate set
+/// can be skipped after a rejection.
```
Nagaprasadvr: a one-liner comment would be nice here

crates/yellowstone-grpc-source/src/lib.rs
```diff
@@ -149,16 +152,115 @@ pub struct YellowstoneGrpcSource {
...
     fn new(config: Self::Config, filters: Filters) -> Self { Self { config, filters } }
 
+    fn supports_filter_updates() -> bool { true }
+
     async fn connect(
         &self,
         tx: Sender<Result<SubscribeUpdate, Status>>,
         status_tx: oneshot::Sender<SourceExitStatus>,
+    ) -> Result<(), ShipsternError> {
+        let (_, closed) = mpsc::channel(1);
```
Nagaprasadvr: i think we dont need to do this here , we can use the channel as an optional argument to self.run

crates/yellowstone-grpc-source/src/lib.rs
```diff
@@ -149,16 +152,115 @@ pub struct YellowstoneGrpcSource {
...
+
+        self.run(tx, status_tx, closed).await
+    }
+
+    async fn connect_with_filter_updates(
+        &self,
+        tx: Sender<Result<SubscribeUpdate, Status>>,
```
Nagaprasadvr: we can return the sender here instead of letting the client pass this as an argument

crates/runtime/src/builder.rs
senzenn: Taken in 9af3e2d. Agreed it doesn't belong there, the const is private with a single use site so nothing renders that block for anyone anyway.

I didn't drop the second paragraph outright though. The part about a newer set replacing a held one instead of queueing behind it isn't written down anywhere else, and it's something a caller can get bitten by, so it moved onto `Runtime::filter_updates` next to the existing best-effort note. The const keeps just the depth rationale.

crates/yellowstone-grpc-source/src/lib.rs
senzenn: Done in 15451b5. `run` takes `Option<Receiver<Filters>>`, `connect` passes `None`, and it lets the `accept_filter_updates` flag go too, since "no channel" and "channel closed" are now the same state.

Worth knowing if you touch it later: the obvious spelling does not work.

```rust
update = filter_updates.as_mut().unwrap().recv(), if accept_filter_updates => {
```

That compiles and then panics on the `connect` path. `select!` evaluates a disabled branch's expression and only skips polling it (step 2 in the tokio docs), so the unwrap still runs with the branch off. I wrote a throwaway test to confirm it rather than trust the docs.

So it sits behind a small helper with the reason next to it:

```rust
async fn next_filter_update(filter_updates: &mut Option<Receiver<Filters>>) -> Option<Filters> {
    match filter_updates {
        Some(rx) => rx.recv().await,
        None => std::future::pending().await,
    }
}
```

The arm in the loop stays a single line.

crates/runtime/src/sources.rs
```diff
@@ -45,4 +48,33 @@ pub trait SourceTrait: std::fmt::Debug + Send + Sync + 'static {
...
+    /// The default ignores `filter_updates` and defers to [`Self::connect`],
+    /// so a source that cannot change its subscription mid-stream needs no
+    /// implementation. Override this together with
+    /// [`Self::supports_filter_updates`].
+    ///
+    async fn connect_with_filter_updates(
+        &self,
+        tx: Sender<Result<SubscribeUpdate, tonic::Status>>,
+        status_tx: oneshot::Sender<SourceExitStatus>,
+        filter_updates: Receiver<Filters>,
```
Nagaprasadvr:
```suggestion
        filter_updates_rx: mpsc::Receiver<Filters>,
```

crates/yellowstone-grpc-source/src/lib.rs
```diff
@@ -199,37 +306,99 @@ impl SourceTrait for YellowstoneGrpcSource {
...
         tracing::debug!("gRPC stream started");
 
+        let mut pending_filters: Option<Filters> = None;
+        let mut filter_updates_sent: u64 = 0;
+
+        // A held set cannot wait on the stream to produce again. The filter it
+        // is replacing may match nothing, and the client swallows its own
+        // keepalive messages rather than yielding them, so there are live
+        // connections on which no update ever arrives to retry from.
+        let mut retry = tokio::time::interval(RETRY_HELD_FILTERS_EVERY);
```
Nagaprasadvr, line 384: im wondering if we need this at all , we can just have a select arm listening from the channel and send it to yellowstone as soon as we receive it

crates/runtime/src/sources.rs
senzenn: Yeah, fair. Renamed in 0a34164, though as `Receiver<Filters>` rather than `mpsc::Receiver` since this file imports `Receiver` directly and `mpsc` isn't in scope here, so the suggestion as-is wouldn't build.

Took the name through the rest of it too (the grpc impl, the mock, the local in `run_impl`) so it's one name end to end.

crates/yellowstone-grpc-source/src/lib.rs
senzenn: Half yes. The send-on-receive arm is already there and does exactly that, the retry only exists for the case where the send fails.

It can fail, and it does recover. `TonicGrpcConnector::connect` swaps a fresh sender into the same `SubscribeRequestSink` after it resubscribes (reconnect.rs:331), so during a reconnect window the send fails, the stream yields nothing, and a retry afterwards lands. Without the hold that update is gone and the subscription quietly keeps the old filters for the rest of the run. Auto-reconnect is on by default so that's the common path.

Retrying off stream traffic instead was the first version and it didn't work: the filter you're replacing may match nothing and the client eats its own keepalives, so there are live connections that never yield anything to retry from. That's why it's a timer.

Where you're right is with auto-reconnect off. No connector, nothing ever swaps a sender back in, so the sink is dead for the rest of the run and the old code still re-sent every 5s forever, logging "holding it until the stream recovers" each time. Fixed in f7c0d7a: drop it and warn once in that mode, keep the hold where it can actually succeed.

crates/runtime/src/handle.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// the source connects. A short burst is buffered, and a sender that gets
+    /// far ahead of the source waits here until it catches up.
+    ///
+    /// # Errors
+    ///
+    /// [`FilterUpdateError::Unsupported`] when the source cannot change its
+    /// subscription mid-stream, and [`FilterUpdateError::Closed`] once the
+    /// runtime has stopped.
+    ///
+    pub async fn send_filter_update(&self, filters: Filters) -> Result<(), FilterUpdateError> {
```
kespinola: How can we access the current registered filters? There should be symmantics for return a copy of the current filters, patching it, and then calling send_filter_update.

crates/runtime/src/runtime_tests.rs
```diff
@@ -373,6 +382,144 @@ impl Handler<SlotUpdate, SlotUpdate> for SlowSlotHandler {
...
+            ids.sort();
+
+            RECEIVED_FILTER_IDS.lock().unwrap().extend(ids);
+        }
+
+        signal_stream_ended(status_tx);
+        hold_channel_open_briefly().await;
+        drop(tx);
+        Ok(())
+    }
```
kespinola, line 444: Is there a way to compose filter update supported sources? 

Right now it only applicable for grpc (which will likely always be the case) but yet every source trait scope is expanded.

Need to look into typing options for such a case. I would like to have something like SourceTrait + FilterUpdateTrait

crates/runtime/src/runtime_tests.rs
```diff
@@ -373,6 +382,144 @@ impl Handler<SlotUpdate, SlotUpdate> for SlowSlotHandler {
...
+const SWAPPED_PARSER_ID: &str = "test::SwappedParser";
+
+#[derive(Debug)]
+struct MockFilterUpdateSource;
+
+#[async_trait]
+impl SourceTrait for MockFilterUpdateSource {
+    type Config = NullConfig;
+
+    fn new(_: NullConfig, _: Filters) -> Self { Self }
```
kespinola: This is always bugged me. Do you know if we can eliminate the implementing SourceTrait have to define this basically stub `new`?

crates/runtime/src/handle.rs
senzenn: This might be a simple approach:
```rs 
pub fn current_filters(&self) -> Option<Filters> {
    self.current_filters.clone()
}

Then callers do:

if let Some(mut f) = handle.current_filters() {
    f.insert(new_id, new_prefilter);
    handle.send_filter_update(f).await?;
}
```
Transparent, no magic, caller owns the race window. Can add closure-based atomic helpers later if needed.

crates/runtime/src/runtime_tests.rs
senzenn: Yeah, agreed. The defaults keep the overhead low right now other sources just inherit `connect_with_filter_updates`, which defers to `connect`, so nothing changes for them. But you're right that it expands the trait surface for something only gRPC uses, and a separate `FilterUpdateTrait` would draw the capability boundary properly.

I left it as-is for now since gRPC is the only source that needs it, and a bound like `SourceTrait + FilterUpdateTrait` adds real complexity for a single implementor. If another provider ever grows mid-stream support, that's the point where splitting it out actually earns its keep.I might add a follow-up so we don't lose the thread.

crates/runtime/src/runtime_tests.rs
senzenn: Yeah, it's annoying boilerplate. The issue is new() takes a generic Config type, so a blanket default won't work across all sources.

Test mocks are the real pain pointcould do a macro to generate them. Real sources actually need custom logic though. Probably worth a follow-up macro for test scaffolding, but refactoring the trait itself would be overkill for now.

crates/runtime/src/handle.rs
kespinola: Addressed in #305, stacked on this branch.

Every handle now shares one view of the filters. `handle.filters()` returns the last set handed to the source, seeded from the registered pipelines, and `handle.update_filters(|f| f.merge(parser.id(), extra))` edits a copy under a lock and publishes the result, so concurrent edits from two handles serialise rather than lose one another. `reset_filters()` goes back to the set the pipelines were built with. `Filters` gained `get`, `insert`, `merge` and `remove` keyed by parser ID for building the next set, and a set naming a parser with no registered pipeline is refused with `FilterUpdateError::UnknownParser` before anything is sent.

One semantic to be aware of: `filters()` means "last set handed off", not "what the server is serving", since the source reports nothing back.

The handle also went synchronous on a `watch` slot in that PR. Details and the behavioural trade-offs are in the #305 description.

senzenn: Resolving. Two things shifted since your first comment, both worth knowing.

The docs now steer read-modify-write to `update_filters` rather than `filters()` -> patch -> `send_filter_update`. `filters()` reads outside the update lock, so two callers doing that lose one of the two edits. `send_filter_update` is documented for sets built from scratch. The README example used to demonstrate the lossy pattern; it does not any more.

Instruction parsers are also not addressable individually. The builder bundles all of them behind one `InstructionPipeline`, so the set carries a single entry under `InstructionPipeline::ID`, and naming an individual instruction parser is refused as `UnknownParser`. The doc example on this file did exactly that. Documented and tested now.

One limit while we are here: `UnknownParser` checks the parser name but not the prefilter kind, so an account prefilter merged onto a slot parser still reaches the wire and every update is discarded. Filed as a follow-up rather than fixed here.

crates/runtime/src/runtime_tests.rs
senzenn: Resolving, since #307 did this.

`Runtime::handle` is gated on `FilterUpdateSource` now, so taking a handle from a source that cannot change its subscription is a compile error rather than a runtime `Unsupported`, and that error variant is gone.

Worth recording why the other half could not move, so nobody retries it: `connect_with_filter_updates` has to stay on `SourceTrait` with a default, because `run_impl` is generic over `S: SourceTrait` and cannot pick a method based on which other traits `Self` implements. Defining `run` on both `impl<S: SourceTrait>` and `impl<S: SourceTrait + FilterUpdateSource>` is `E0592`. I checked that with rustc rather than assuming it.

One residual, filed as a follow-up rather than held here: the marker is empty and the default discards the receiver, so a source can `impl FilterUpdateSource for X {}` without overriding and get a working handle whose every update returns `Ok(())` and reaches nothing. Since #307 removed `Unsupported`, that case is now silent where it used to be reported. Nothing in-tree hits it. Folding the receiver into `connect` as a required parameter would close it, and #308 already moves `connect` in that direction.

crates/runtime/src/runtime_tests.rs
senzenn: Resolving, pointing at #308.

You were right and my answer on this thread was wrong. I said real sources need custom construction logic. None do: all six were the same struct literal storing config and filters, and #308 removes `Config` and `new` from `SourceTrait` outright.

Review notes are on that PR. The one thing worth repeating here is that the compatibility note undersells it: moving `Config` onto `FromConfig` changes `ShipsternConfig<S::Config>`, so downstream generic code bounded `S: SourceTrait` that calls `try_build` stops compiling. That is a break for callers, not just implementors.

PR #315, opened 2026-09-09
feat(deps): move to the Agave 4.2.2 stack and handle transaction v1

### LINKED ISSUE #310: Upgrade the Solana/Agave dependency stack from 3.x to 4.2
<html><head></head><body><h2>Problem</h2><p>We're still pinned to Agave/Solana <code inline="">3.x</code>, while the clusters we're reading from are already on <code inline="">4.x</code>.</p><p>Current versions:</p>
cluster | version
-- | --
mainnet-beta | 4.2.2
devnet | 4.3.0-beta.3
testnet | 4.3.0-beta.3

<p>This is starting to cause real problems. In particular, <code inline="">VersionedMessage::V1</code> is only available in <code inline="">solana-message</code> 4.x, so the RPC, snapshot, mock, and parser paths can't handle V1 transactions.</p><p>Devnet already has real V1 traffic.</p><h2>Proposed change</h2><p>Upgrade the workspace to Agave/Solana <code inline="">4.2.x</code>.</p><p>This also means:</p><ul><li><p><code inline="">spl-token-2022</code> 10 → 11</p></li><li><p><code inline="">solana-loader-v3-interface</code> 6 → 7</p></li><li><p>MSRV 1.90 → 1.93</p></li><li><p>enable <code inline="">wincode</code> for <code inline="">solana-message</code> and <code inline="">solana-transaction</code></p></li><li><p>update the affected Agave APIs</p></li><li><p>add V1 handling and preserve <code inline="">Message.config</code></p></li></ul><p>I'm targeting <strong>4.2.2</strong>, not 4.3, since 4.2.2 is what mainnet-beta currently runs.</p><h2>One important gotcha</h2><p>Adding the <code inline="">V1</code> match arm isn't enough.</p><p><code inline="">jetstreamer-source</code> currently sets:</p><pre><code class="language-rust">config: None
</code></pre><p>That won't necessarily fail compilation after adding V1, but it would silently drop the V1 inline budget.</p><p>We should populate it from the V1 message.</p><h2>Jetstreamer</h2><p>The previous dependency conflict with <code inline="">jetstreamer</code> is no longer blocking this. Its main branch has already moved to Agave 4.2.</p><p>The remaining release issue is that the Agave 4.2 version hasn't been published to crates.io yet.</p><h2>Testing</h2><p>So far:</p><ul><li><p>workspace tests pass on Rust 1.93</p></li><li><p>clippy passes</p></li><li><p>4.2.2 dependency graph resolves</p></li><li><p>real V1 transactions have been tested from devnet</p></li><li><p><code inline="">wincode</code> round-trips the V1 transactions correctly</p></li></ul><p>The <code inline="">jetstreamer-source</code> port still needs to be finished.</p><h2>Related</h2><ul><li><p>#209: MSRV</p></li><li><p>#304: V1 inline budget</p></li><li><p>#308: <code inline="">SourceTrait</code> changes</p></li><li><p>#311: V1 support</p></li></ul><p>I'd split the work into dependency/API changes first, then the <code inline="">jetstreamer-source</code> port, then the release.</p></body></html>

senzenn:
Bump Shipstern to the Agave 4.2.2 stack and add V1 handling to the Jetstreamer transaction conversion.

Closes #310.

Legacy and V0 stay on the existing path. V1 preserves `Message.config` through the protobuf and into `InstructionShared::transaction_config`, where parsers can read it.

The Agave 4.2 reward API changed as well. `RewardInfo::commission` is gone, so `commission_bps` is forwarded exactly, and `DeactivatedStake` is handled explicitly. Token-2022 goes 10 to 11 for the new extension types, including PermissionedBurn.

### The part worth reviewing

`convert::transaction` matched only Legacy and V0 and hardcoded `config: None`. Adding a `VersionedMessage::V1` arm makes it compile, but that line compiles too, and it drops the inline budget on every V1 transaction. Since `versioned` is `true` for both V0 and V1, the two also stop being distinguishable.

The proto marks a message as V1 by the presence of `config`, not its contents:

```protobuf
// Set only for V1 transaction messages (SIMD-0385). Absent for legacy/V0.
optional TransactionConfig config = 7;
```

So the V1 arm wraps it unconditionally. A V1 message with no budget fields set still has to come out as `Some(TransactionConfig::default())`, or it reads as V0 downstream. The three version-dependent fields moved into a local `MessageParts` struct so every arm states all three:

```
Legacy  versioned false  config None      lookups []
V0      versioned true   config None      lookups preserved
V1      versioned true   config Some(..)  lookups []
```

### Rewards

`RewardInfo` lost `commission` in 4.2 and keeps only `commission_bps`, so the old `bps = commission * 100` has no input. Basis points forward verbatim now.

The percent field is still populated, but only where the bps divide evenly. I nearly dropped it, since Agave emits either the percent or the bps depending on the SIMD-0291 flag and never both. It stays because `shipstern-block-meta-parser` puts `commission` on its own public `Reward` and has no `commission_bps` to fall back on, so blanking it would silently lose the commission for block-meta consumers. 1234 bps has no exact `u8`, so those report empty rather than a truncated 12.

`solana_transaction_status::Reward` kept both fields, so on that path each forwards from its own source.

### Tests

Legacy and V0 conversion is covered by the existing tests plus three real mainnet transactions, one carrying an address table lookup. V1 is covered by four real devnet fixtures and wincode round-trips.

| fixture | slot | bytes | config mask |
| --- | --- | --- | --- |
| `v1_all_config_fields` | 495522053 | 1296 | `0x1f` all four |
| `v1_large` | 495521680 | 3276 | `0xc` |
| `v1_over_1232` | 495525717 | 2279 | `0xc` |
| `v1_small` | 495524338 | 196 | `0xc` |

Three are above the old 1232 byte boundary. They go through the same `wincode::deserialize` call Jetstreamer uses (`Transaction::as_parsed`, `jetstreamer-firehose/src/transaction.rs:113`), then `convert::transaction()`, then into `InstructionUpdate::build_from_txn`, and the config is compared against what came off the wire. Message bytes and signatures are checked by re-serializing.

Re-checked against live mainnet after `enable_tx_v1` activated there, which happened after this merged. 1493 V1 transactions from slots 447283712-447283736 went through the same `wincode::deserialize` -> `convert::transaction` -> `InstructionUpdate::build_from_txn` path as the fixtures above. All 1493 decoded as V1, kept their config through the proto, and reached `InstructionShared` intact. 2952 instructions built, zero failures. 577 of them are over the old 1232-byte cap, the largest 2399 bytes, and the observed config masks run from two fields set up to all four.

Testnet was sampled the same way and agrees: 30 transactions over 5 slots, all clean, though that window held only small transactions with two config fields set.

bincode is not the V1 wire format. `solana-message` says so in its own V1 `Serialize` arm and `solana-transaction`'s tests say the same, so there's a test pinning it in case someone rewrites the round-trip later.

Malformed input covers every truncation, unknown config mask bits, half-set priority fee bits, and a V1 message behind a legacy signature count. All fail to decode. 4097 bytes is in the size sweep on purpose: `MAX_TRANSACTION_SIZE` is the validator's limit, not the codec's, and Shipstern holds no transaction size constant of its own.

### Notes if you consume V1

- The compute budget is essentially no longer an instruction. Code that reads the priority fee or CU limit by scanning for ComputeBudget program instructions gets nothing on almost every V1 transaction. Two mainnet samples found 0 of 1493 and 2 of 2940 V1 transactions carrying the ComputeBudget program key, against a large majority of V0 and legacy in the same blocks. It is not literally never, so treat it as a fallback at best and read `InstructionShared::transaction_config` as the source of truth.

- `transaction_config.is_some()` means the transaction is V1, not that it set a budget. The proto marks V1 by the presence of the field rather than its contents, so by construction a V1 message with every field unset still serializes as `Some(TransactionConfig::default())`. Treat it as the version discriminator and read the individual fields for actual values.

- `versioned` is true for both V0 and V1, so it cannot tell them apart. `config` is the only discriminator.

- V1 carries no address table lookups. `address_table_lookups` is empty on every V1 message, including all 1493 in the mainnet sample.

- V1 transactions exceed the old 1232-byte cap. 577 of those 1493 were larger, up to 2399 bytes. `MAX_TRANSACTION_SIZE` is the validator's limit, not the codec's, and Shipstern holds no transaction size constant of its own.

### What isn't covered

At merge time mainnet V1 was not active, so the live mainnet checks here covered Legacy and V0 only and the V1 fixtures came from devnet. `enable_tx_v1` has since activated on mainnet-beta, and the mainnet result above is from after that.

Snapshot compatibility against a real Agave 4.2 archive was not tested. The source compiles and its tests pass, which isn't the same thing.

No in-tree parser reads `transaction_config` yet.

PermissionedBurn wasn't found on mainnet, but that's a bounded sample of roughly 3,850 Token-2022 accounts with extensions, not a survey of the program. `NativeExtensionData` still returns `Err` for it because there's no proto message for `PermissionedBurnConfig`, but nothing calls that conversion.

### Other changes

`crates/mock/src/lib.rs` asked for `maxSupportedTransactionVersion: 0`, set when V0 was the newest version. That's the node's ceiling, so a V1 fixture could never be fetched through it. Raised to 1, which returns identical JSON for legacy and V0 on a 4.2.2 node.

`wincode` is named explicitly on `solana-message` and `solana-transaction` now. V1 only serializes through it and it was arriving transitively.

Depends on anza-xyz/jetstreamer#95 (merged). `jetstreamer-firehose` is still a git dependency, so `cargo publish` stays blocked until they cut a release.


crates/jetstreamer-source/src/lib.rs
```diff
@@ -1755,64 +2787,114 @@ mod convert {
...
+    /// V1     -> versioned: true,  address_table_lookups: [],     config: Some(..)
+    /// ```
+    ///
+    struct MessageParts {
+        header: solana_message::MessageHeader,
+        account_keys: Vec<solana_pubkey::Pubkey>,
+        /// Legacy and V0 call this `recent_blockhash`, V1 calls it
+        /// `lifetime_specifier`. Both are the same 32 bytes and the proto
+        /// carries them in the same field.
+        lifetime: Vec<u8>,
```
Nagaprasadvr: can use [u8;32], can also define a type `Lifetime`

Cargo.toml
Nagaprasadvr, line 31: can remove bincode if not used ? we should use wincode everywhere as we are migrating to 4.x.x

Nagaprasadvr (changes requested):
looks good few things to address

crates/jetstreamer-source/src/lib.rs
senzenn: Made it `Hash`, both `recent_blockhash` and `lifetime_specifier` already are one so the `Vec` was just throwing that away. `solana_message` re-exports it so no new dep. Skipped the newtype since it wouldn't add anything `Hash` doesn't already give.

Cargo.toml
senzenn: Switched to wincode, agave's own storage-proto moved that same field over in 4.2 and we were still writing the 3.x convention. Bytes come out identical for `TransactionError` so nothing changes on the wire, added a round-trip test since nothing covered that field. Can't remove bincode though, `bpf-loader-parser` needs it and `solana-loader-v3-interface` 7 has no wincode feature, so it stays as a dev-dep here.

Nagaprasadvr (approved):

PR #321, opened 2026-09-13
fix: emit generated integer literals unsuffixed

senzenn:
quote! interpolates a bare integer as a suffixed literal, so a count, size or offset
reaches the generated parser as 558usize rather than 558. That pins the emitted code to
whatever width codama gives the field today, and codama already changed those types once
in 0.9.3, a patch release. When a count goes usize to u64, every generated parser stops
compiling:

    error[E0308]: mismatched types
       |     data.len() == 558u64
       |                   ^^^^^^ expected `usize`, found `u64`

The IDL never changed. Only the width of a codama field did.

utils::unsuffixed emits the literal with no suffix and lets inference pick the width at
the use site, which is usize in every case here since the generated code indexes data and
declares usize consts. 20 call sites across account_parser.rs, instruction_parser.rs and
shipstern_parser.rs.

generated_parsers_contain_no_width_pinned_literals expands all 31 fixture IDLs and asserts
no width-suffixed literal appears. On main it reports 287 width-pinned literals across 31
IDLs; here it reports 0. Two things keep that from being vacuous: forcing one site back to
a suffixed literal fails to compile with

    error[E0277]: the type `[KeyBytes<32>]` cannot be indexed by `std::ops::RangeFrom<u64>`

so inference at these sites is load-bearing, and scanner_separates_literals_from_type_mentions
pins the scanner (558usize hits, `const N: usize` does not, swap_v2usize does not).

u32 is outside the guard on purpose. prost field tags are u32 by the protobuf spec and are
numbered from field order, so under the proto feature the generated code carries tens of
thousands of legitimately suffixed u32 literals.

Works on codama 0.9.1, so it doesn't depend on the version bump.

Tested: 363 passed, 0 failed. clippy -Dwarnings and fmt --check clean.

Based on #320, which syncs the tests/proc-macro-events lockfile. That branch is a shared
prerequisite; the fixture-guard stack needs it too.


kespinola (commented):
Is there any impact on the proto feature?

senzenn:
Yeah, 3 of the 20 sites are actually under the proto gate, the dispatch index
consts in shipstern_parser.rs. They're fine though, each one expands into
`pub const X: Option<usize> = Some(..)`, so the annotation pins usize and
inference takes it from there.

Ran the guard with --features proto, the proc-macro-test crate, and
shipstern-proto --all-features (the dispatch_indices_match_proto tests resolve
the const against the real descriptor). All green. Nothing that emits prost
types got touched.

u32 being out of the guard isn't arbitrary either, adding it back fails under
proto with 527 hits, all prost field tags.

kespinola (approved):

PR #325, opened 2026-09-17
fix(proc-macro): honor Codama byte length prefixes

Aursen:
Codama `sizePrefixTypeNode(bytes, u64)` currently becomes an ordinary `Vec<u8>` with Borsh's u32 length prefix. This silently shifts the decoded payload by four bytes and drops its final four bytes.

The regression uses a real ProtocolCPI instruction from mainnet slot 447727788:

```text
07 1000000000000000 b7cc0e45114cdf69 0100000000000000
```

The u64 length is 16. Before this fix, the decoded bytes are `00000000b7cc0e45114cdf6901000000`; afterward they are the complete inner instruction, `b7cc0e45114cdf690100000000000000`. The regression was run against unmodified main and failed with exactly that mismatch.

Preserve the declared prefix format and endianness in the IR and generate matching Borsh field codecs. Rust fields remain `Vec<u8>` and protobuf fields remain `bytes` with unchanged tags. Native u32-LE encoding stays on the existing path. The codecs handle unsigned fixed-width and compact-u16 prefixes, including optional/repeated/fixed-array byte fields, and reject truncation and length overflow without allocating from an untrusted declared length.

Validation:

- `cargo +nightly-2026-02-25 fmt --all -- --check`
- `cargo test --locked --workspace --lib --tests`: 387 passed, 6 ignored.
- `cargo clippy --locked --all-targets --tests --no-deps --features shipstern-kafka-sink/experimental-account-parser -- -Dwarnings`
- `cargo test --locked --manifest-path tests/proc-macro-events/Cargo.toml --tests`: 45 passed.
- `cargo test --locked -p shipstern-proto --lib --features parser,stream`: 5 passed.
- `cargo test --locked -p shipstern-kafka-sink --features experimental-account-parser --lib --tests`: 36 passed.
- An isolated consumer without the proc-macro's `proto` feature also compiled and passed a Borsh round-trip test.

The regression checks the complete CPI payload, protobuf wire compatibility, prefix widths/endianness, compact-u16 boundaries, empty/truncated/oversized input, and composition with options and arrays. No dependency or lockfile changes.

Downstream follow-up: Janus must rebuild/deploy with a release containing this fix. Existing damaged parsed records need reparsing from original transactions; replaying the already-truncated payload cannot recover the missing bytes. This PR fixes the codec and does not perform that deployment or replay.


crates/proc-macro/src/intermediate_representation/helpers.rs
```diff
@@ -347,12 +347,22 @@ fn materialize_tuple_message(
...
+                if prefix.format == NF::U32 && prefix.endian == Endianness::Le {
+                    FieldTypeIr::Scalar(ScalarIr::Bytes)
+                } else {
+                    FieldTypeIr::Scalar(ScalarIr::SizePrefixedBytes {
+                        format: prefix.format,
+                        endian: prefix.endian,
+                    })
+                }
+            },
             _ => FieldTypeIr::Scalar(ScalarIr::String),
```
senzenn: Same bug, one arm down. Every non-bytes size prefix lands on `ScalarIr::String`, and `String`'s borsh codec reads a u32 LE prefix, so `sizePrefix(string, u8)` or `sizePrefix(string, u64)` shifts and truncates exactly the way the bytes case did.

Nothing in this repo hits it (I checked: all ~30 `sizePrefix` nodes in `tests/idls` are `u32`/`le`), but the codec you just wrote is a `String::from_utf8` away from covering it. Can you extend it here, or open an issue, so we don't find the string case the same way we found this one?

crates/proc-macro/src/render/size_prefixed_bytes.rs
```diff
@@ -0,0 +1,180 @@
...
+        Span::call_site(),
+    );
+    let write = LitStr::new(
+        &format!("Self::__shipstern_write_{}", field.name),
+        Span::call_site(),
+    );
+    quote! { #[borsh(deserialize_with = #read, serialize_with = #write)] }
+}
+
+pub(super) fn methods(field: &FieldIr, in_module: bool) -> TokenStream {
```
senzenn: This emits a read/write pair per field with the codec body inlined into each one. Everywhere else in the generated output, a non-native borsh encoding is a generic helper emitted once per program module and called by turbofish: `borsh_deserialize_option_prefix::<N, ONE, BE, _>`, `borsh_deserialize_fixed_array::<T, N, _>`, `borsh_deserialize_fixed_bytes::<N, _>`. You already reuse the option-tag helpers a few lines down, so it's only the bytes codec that goes its own way.

Could this be `borsh_{de,}serialize_size_prefixed_bytes::<N, BE, _>` plus a shortU16 variant, living next to the others in `option_borsh_helpers`? Keeps the generated module readable and stops the body being duplicated per field.

crates/proc-macro/src/render/size_prefixed_bytes.rs
```diff
@@ -0,0 +1,180 @@
...
+                        quote!(#parent borsh_serialize_option_prefix::<#byte_len, #one_value, #big_endian, _>(value.is_some(), writer)?;),
+                    )
+                },
+                OptionPrefixIr::ShortU16 => (
+                    quote!(#parent borsh_deserialize_short_u16_option_prefix(reader)?),
+                    quote!(#parent borsh_serialize_short_u16_option_prefix(value.is_some(), writer)?;),
+                ),
+            };
+            // Variable-size byte strings cannot be the item of a fixed Codama option.
+            debug_assert!(encoding.none_padding.is_none());
```
senzenn: Make this a hard `panic!`.

It's unreachable today: `build_option_encoding` already panics when a fixed option has a variable-size item, and `fixed_size_of_type` returns `None` for `SizePrefix`, so I'm not worried about it firing. But `debug_assert` compiles out in a release-profile proc-macro, so the day it does become reachable the failure mode is a codec that silently skips the `None` padding, which is the same class of silent wire corruption this PR exists to fix. Cheap to make it loud.

crates/proc-macro/src/render/size_prefixed_bytes.rs
```diff
@@ -0,0 +1,180 @@
...
+                quote! { #write_tag if let Some(value) = value { write_bytes(value, writer)?; } Ok(()) },
+            )
+        },
+        LabelIr::Repeated | LabelIr::FixedArray(_) => {
+            let (count, write_count) = match &field.label {
+                LabelIr::FixedArray(n) => {
+                    let n = crate::utils::unsuffixed(*n as u64);
+                    (quote!(#n), quote! {
+                        if value.len() != #n {
+                            return Err(::borsh::io::Error::new(::borsh::io::ErrorKind::InvalidInput, "invalid fixed byte-array count"));
```
senzenn: Worth flagging that this is stricter than the existing path: `borsh_serialize_fixed_array` in `shipstern_parser.rs` writes however many elements the Vec holds with no length check.

I think your version is the correct one and the old helper is the bug, but we now have two fixed-array behaviours in the same generated module. Fine to leave as-is and align the old one separately; I mainly want it written down rather than discovered later.

crates/proc-macro/src/render/size_prefixed_bytes.rs
```diff
@@ -0,0 +1,180 @@
...
+                #write_length
+                writer.write_all(value)
+            }
+            #write
+        }
+    }
+}
+
+fn prefix_codec(format: NumberFormat, endian: Endianness) -> (TokenStream, TokenStream) {
+    if format == NumberFormat::ShortU16 {
```
senzenn: The format check runs before `endian` is looked at, so `numberTypeNode { format: shortU16, endian: be }` is silently decoded as LE. Defensible, since shortU16 has no endianness to speak of, but the arm below panics on formats it can't honour, so rejecting `be` here would be more consistent than quietly ignoring it.

crates/proc-macro/src/render/size_prefixed_bytes.rs
```diff
@@ -0,0 +1,180 @@
...
+            },
+        );
+    }
+    let (ty, size) = match format {
+        NumberFormat::U8 => (quote!(u8), 1),
+        NumberFormat::U16 => (quote!(u16), 2),
+        NumberFormat::U32 => (quote!(u32), 4),
+        NumberFormat::U64 => (quote!(u64), 8),
+        NumberFormat::U128 => (quote!(u128), 16),
+        other => panic!("unsupported byte length prefix: {other:?}; expected an unsigned integer"),
```
senzenn: This doesn't say which field or struct tripped it. I checked what a consumer actually sees and it's a bare `proc macro panicked` plus `help: message: unsupported byte length prefix: I64; expected an unsigned integer`, with no other context. Threading `field.name` through would save someone grepping an IDL by hand.

tests/proc-macro/tests/size_prefixed_bytes.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+    assert!(borsh::to_vec(&ShortU16LeBytes {
+        data: vec![0; 65_536],
+        sentinel: 0
+    })
+    .is_err());
+}
+
+#[test]
+fn composes_with_option_and_array_encodings() {
```
senzenn, line 103: `Containers` is a defined type, so it renders at the top level and the option helpers resolve with no path prefix. The `in_module` branch of `methods()` emits `super::borsh_deserialize_option_prefix`, and nothing here reaches it, because `ProtocolCpi`'s only size-prefixed argument is singular.

I built an IDL with `option(sizePrefix(bytes, u64))` and a prefixed array as instruction arguments and it round-trips byte-exact, so the path is correct, just untested. Moving one of these cases onto an instruction argument would cover it.

senzenn (changes requested):
Verified this end to end rather than just reading it, and the fix is right.

I reverted to 5766c10 with your test and IDL dropped in: all four fail there, and `data` comes back as `00000000b7cc0e45114cdf6901000000`, which is exactly the shift and the four lost bytes you describe. On the branch all four pass. `fmt` and `clippy -Dwarnings` are clean, the workspace is 387 passed / 6 ignored, the events workspace passes. I also scanned every IDL in the repo: all of the `sizePrefixTypeNode` nodes here are `u32`/`le`, so no bundled parser's generated output moves. That's a good property for a fix of this shape.

The test is the strongest part of the PR. Real mainnet payload, byte-exact wire assertions, the full width and endian matrix, truncation swept over every prefix length, shortU16 non-minimal rejection plus the boundary table, and it actually demonstrates that a `u64::MAX` declared length doesn't preallocate.

One thing blocks merge: there is no CHANGELOG entry. Everything since #309 adds one under `[Unreleased]` with the PR link and author, and this change needs it more than most, since decoded output changes for affected IDLs and the reparse consequence you wrote in the description belongs somewhere consumers will actually read. Please add it under `### Fixed`, including the point that replaying an already-truncated payload cannot recover the missing bytes.

Two more I would like settled, both inline: the same bug still exists for non-u32 prefixed strings, and the codec's per-field shape diverges from the helper convention used for every other non-native borsh encoding here. The rest are nits.

Aursen:
Hi @senzenn can you check again?

senzenn:
Hey @Aursen thank you for your contribution! Reviewing it now.

senzenn (approved):
All good on my end, everything from the last round is in.

Didn't just read it this time. Ran the paths no fixture covers: the account `try_unpack` path with a size-prefixed field, the `short_u16_opt_size_prefixed` pair (no fixture emits it, so CI never compiles it), and a full label x variant prost matrix checking `encoded_len` against what `encode_to_vec` actually writes. Also diffed the shortU16 codec against `solana-short-vec`, which agrees on every length across the width boundaries and on all 65,536 two-byte prefixes. Fuzzed it with 40k random inputs plus single-byte mutations of a valid encoding, no panics in debug or release.

One thing that's my fault, not yours: the comment at `size_prefixed_bytes.rs:55` cites `build_option_encoding`. No such function, it's `option_encoding` at `helpers.rs:218`. I handed you the wrong name in my review. Not worth a respin, I'll fix it separately.

The array-under-`sizePrefix` case from my `helpers.rs` comment is still half open, but that's pre-existing on main and not yours to close here. Filing a follow-up for it and for the Codama wrapper handling.

Thanks for the quick turnaround.

crates/proc-macro/src/render/size_prefixed_bytes.rs
senzenn: Yep, that's what I was after.

Small addition for whoever picks up the alignment: it's four helpers, not one. `borsh_serialize_fixed_array`, `borsh_serialize_fixed_array_fixed_bytes`, `borsh_serialize_fixed_array_f32_permissive` and `borsh_serialize_fixed_array_f64_permissive` all write however many elements they're handed. Yours is the one that behaves. Not for this PR.

